PR #1, opened 2022-06-23
Design adjustments

DiChizhik:


Application/SceneDelegate.swift
```diff
@@ -37,9 +37,12 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
             mostPopularTabNavigationController
         ]
         
+//        let launchViewController = LaunchViewController()
```
GMinucci: Now that we have git don't worry about keeping comments. You can always check on the file history on git to recover old code when needed. Of course, if you need to keep the comment, for example when you have a feature that will be done soon you can leave it, otherwise I'd recommend to remove comments to improve code readability :)

Application/SceneDelegate.swift
```diff
@@ -37,9 +37,12 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
...
+        
         window = UIWindow(windowScene: windowScene)
     
         window?.rootViewController = tabBarController
+//        window?.rootViewController = launchViewController
```
GMinucci: We can delete this comment too

Application/SceneDelegate.swift
```diff
@@ -73,9 +76,10 @@ class SceneDelegate: UIResponder, UIWindowSceneDelegate {
     }
 
     func setupTabBar() {
-        UITabBarItem.appearance().setTitleTextAttributes([.foregroundColor: UIColor(named: "grey")], for: .normal)
-        UITabBarItem.appearance().setTitleTextAttributes([.foregroundColor : UIColor(named: "acidGreen")], for: .selected)
-        UITabBar.appearance().tintColor = UIColor(named: "acidGreen")
+        UITabBarItem.appearance().setTitleTextAttributes([.foregroundColor: UIColor(named: "nonSelectedTab")!], for: .normal)
+        UITabBarItem.appearance().setTitleTextAttributes([.foregroundColor : UIColor(named: "selectedTab")!], for: .selected)
+        UITabBar.appearance().tintColor = UIColor(named: "selectedTab")
+        UITabBar.appearance().unselectedItemTintColor = UIColor(named: "nonSelectedTab")
```
GMinucci: Just a note: The appearance proxy will set the default behaviour to the entire app. So in case you have multiple UITabbar with diff styles this needs to be done directly to each tabbar. For your app there is no issue here

Models/Movie.swift
```diff
@@ -7,31 +7,39 @@
...
-        case title
         case voteAverage = "vote_average"
     }
     
     init(from decoder: Decoder) throws {
         let values = try decoder.container(keyedBy: CodingKeys.self)
         id = try values.decode(Int.self, forKey: .id)
-        posterPath = try values.decode(String.self, forKey: .posterPath)
+        
+        let posterPathString = try values.decode(String.self, forKey: .posterPath)
```
GMinucci: From the API docs [here](https://developers.themoviedb.org/3/movies/get-now-playing), `poster_path` can be String or Null so we should use the optional decoding here
```suggestion
        let posterPathString = try values.decodeIfPresent(String.self, forKey: .posterPath)
```

Models/MovieDetails.swift
```diff
@@ -26,10 +26,32 @@ struct MovieDetails: Codable {
         case runtime
         case overview
     }
+    
+    init(from decoder: Decoder) throws {
+        let values = try decoder.container(keyedBy: CodingKeys.self)
+        
+        title = try values.decode(String.self, forKey: .title)
+        
+        let posterPathString = try values.decode(String.self, forKey: .posterPath)
```
GMinucci: Same as previous comment, from [here](https://developers.themoviedb.org/3/movies/get-movie-details) `poster_path` can be null so:
```suggestion
        let posterPathString = try values.decodeIfPresent(String.self, forKey: .posterPath)
```

Models/MovieDetails.swift
```diff
@@ -26,10 +26,32 @@ struct MovieDetails: Codable {
...
+        spokenLanguages = try values.decode([Language].self, forKey: .spokenLanguages)
+        genres = try values.decode([Genre].self, forKey: .genres)
+        
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateStyle = .medium
+        dateFormatter.timeStyle = .none
+        dateFormatter.locale = Locale(identifier: "en_US")
+        releaseDate = dateFormatter.string(from: try values.decode(Date.self, forKey: .releaseDate))
+        
+        let runtimeInt = try values.decode(Int.self, forKey: .runtime)
```
GMinucci: Same as poster path, this property can be null from [here](https://developers.themoviedb.org/3/movies/get-movie-details), so `decodeIfPresent`

Models/MovieDetails.swift
```diff
@@ -26,10 +26,32 @@ struct MovieDetails: Codable {
...
+        runtime = "\(Int(runtimeInt / 60))h \(runtimeInt % 60)min"
+        
+        overview = try values.decode(String.self, forKey: .overview)
```
GMinucci, line 56: This one too, `decodeIfPresent`

Resources/Assets.xcassets/mostRecentIcon.imageset/Contents.json
```diff
@@ -1,15 +1,21 @@
 {
   "images" : [
     {
-      "filename" : "3 Icons_2 Navigation_3 Business_02 Accounts.svg",
-      "idiom" : "universal"
+      "filename" : "mostPopular.svg",
```
GMinucci, line 4: For SVG files we don't need to provide all 3 scales. SVGs are vector files so they don't "loose" quality when resizing. We can the use single scale option.

Services/MovieDataService.swift
```diff
@@ -85,8 +85,11 @@ class MovieDataService: MovieDataServiceProtocol {
                 completion(nil)
                 return
             }
+            let dateFormatter = DateFormatter()
+            dateFormatter.dateFormat = "yyyy-MM-dd"
             let jsonDecoder = JSONDecoder()
-//            jsonDecoder.dateDecodingStrategy = .iso8601
+            jsonDecoder.dateDecodingStrategy = .formatted(dateFormatter)
```
GMinucci, line 99: 👏

ViewControllers/DetailViewController.swift
```diff
@@ -12,17 +12,18 @@ class DetailViewController: UIViewController {
     static var identifier: String{NSStringFromClass(self)}
```
GMinucci, line 12: 👀
This is super nice! In case you wanna read more discussions around how to get a name from a type check this [thread](https://forums.swift.org/t/relying-on-string-describing-to-get-the-name-of-a-type/16391). Long story short, the piece of code you are using is the recommended one

ViewControllers/DetailViewController.swift
```diff
@@ -12,17 +12,18 @@ class DetailViewController: UIViewController {
...
     
     var selectedMovieId: Int?
-    
-    let movieDataService = MovieDataService()
-    
     var movieDetails: MovieDetails?
     var languages = [String]()
+    var languagesTest = ["German", "Dutch", "English", "Chinese", "Icelandic", "Russian", "Spanish", "Italian", "Japanese", ]
```
GMinucci, line 17: I think we are not using this so we can remove it

ViewControllers/DetailViewController.swift
```diff
@@ -12,17 +12,18 @@ class DetailViewController: UIViewController {
...
     var genres = [String]()
+    var genresTest = ["thriller", "action", "science finction", "romcom", "adventure"]
```
GMinucci, line 19: This is not being used as well

ViewControllers/DetailViewController.swift
```diff
@@ -12,17 +12,18 @@ class DetailViewController: UIViewController {
...
+    
+    let movieDataService = MovieDataService()
```
GMinucci, line 21: 👏

ViewControllers/DetailViewController.swift
```diff
@@ -96,6 +97,51 @@ class DetailViewController: UIViewController {
...
+    
+    private func makeViewData() -> [CollectionViewSection] {
+        let languages: [CollectionViewItem] = languages.map{CollectionViewItem(title: $0)}
+        let languagesSection = CollectionViewSection(identifier: CollectionViewSectionIdentifier.languages, items: languages)
+        
+        let genres: [CollectionViewItem] = genres.map{CollectionViewItem(title: $0)}
+        let genreSection = CollectionViewSection(identifier: .genres, items: genres)
+    
+        return [languagesSection, genreSection]
+        }
```
GMinucci, line 142: I loved it! 🔥

ViewControllers/DetailViewController.swift
```diff
@@ -203,29 +238,62 @@ class DetailViewController: UIViewController {
...
-            if let url = movieDataService.getMoviePosterURL(posterPath: path) {
-                imageView.kf.setImage(with: url)
-            }
+        if let path = movieDetails.posterPath {
+            imageView.kf.setImage(with: path)
+        }
         
-        releaseDateLabel.text = movieDetails.releaseDate
-        durationLabel.text = "\(movieDetails.runtime)m"
+        releaseDateLabel.text = "\(movieDetails.releaseDate)"
```
GMinucci: `movieDetails.releaseDate` is a string now so we can:
```suggestion
        releaseDateLabel.text = movieDetails.releaseDate
```

ViewControllers/DetailViewController.swift
```diff
@@ -203,29 +238,62 @@ class DetailViewController: UIViewController {
...
+    func collectionView(_ collectionView: UICollectionView, cellForItemAt indexPath: IndexPath) -> UICollectionViewCell {
+        
+        guard let cell = collectionView.dequeueReusableCell(withReuseIdentifier: DetailCollectionViewCell.reuseIdentifier, for: indexPath) as? DetailCollectionViewCell else { fatalError() }
+        let item = viewData[indexPath.section].items[indexPath.item].title.lowercased().capitalizingFirstLetter()
+        
+        if viewData[indexPath.section].identifier == .languages {
+        cell.configure(title: item, backgroundColor: UIColor(named: "languageBackgroundColor")!, borderColor: UIColor(named: "languageBorderColor")!)
+        } else {
+            cell.configure(title: item, backgroundColor: UIColor(named: "genreBackgroundColor")!, borderColor: UIColor(named: "genreBorderColor")!)
+        }
```
GMinucci: Just a small suggestion, in cases like this I feel that `switch` make it more readable so it would be:
```suggestion
        switch viewData[indexPath.section].identifier {
        case .languages:
            cell.configure(title: item, backgroundColor: UIColor(named: "languageBackgroundColor")!, borderColor: UIColor(named: "languageBorderColor")!)
        
        case .genres:
             cell.configure(title: item, backgroundColor: UIColor(named: "genreBackgroundColor")!, borderColor: UIColor(named: "genreBorderColor")!)
        }
```
Then, in case you add more sections the compiler will remind you to include the config as well

ViewControllers/MostPopularCollectionViewController.swift
```diff
@@ -34,36 +34,27 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
...
-//        override func viewDidAppear(_ animated: Bool) {
-//            super.viewDidAppear(animated)
-//
-//            reviewsScoreIndicator.layer.cornerRadius = reviewsScoreIndicator.bounds.size.width / 2
-//        }
-        reviewsScoreIndicator.layer.cornerRadius = 6
-        reviewsScoreIndicator.layer.masksToBounds = true
-        reviewsScoreIndicator.backgroundColor = UIColor(named: "red")
+        reviewsScoreIndicator.heightAnchor.constraint(equalToConstant: 15).isActive = true
+        reviewsScoreIndicator.widthAnchor.constraint(equalTo: reviewsScoreIndicator.heightAnchor, multiplier: 1.13).isActive = true
```
GMinucci, line 57: 👏

ViewControllers/MostPopularCollectionViewController.swift
```diff
@@ -91,16 +82,34 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
...
     }()
     
     override func viewDidLoad() {
         super.viewDidLoad()
         
-        title = "Most popular"
-        navigationController?.navigationBar.titleTextAttributes = [.foregroundColor: UIColor.white, .font: UIFont.systemFont(ofSize: 24, weight: .heavy)]
+        setupUI()
+        
+        movieDataService.getMostPopularMoviesList { [weak self] (moviesList: [Movie]?) in
```
GMinucci: Just a bit of syntactic sugar, we don't need to specify the type in the closure and can leave to the swift's automatic type inference
```suggestion
        movieDataService.getMostPopularMoviesList { [weak self] moviesList in
```

ViewControllers/MostPopularCollectionViewController.swift
```diff
@@ -91,16 +82,34 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
...
+            if let moviesList = moviesList {
+                self?.movies = moviesList
+            }
```
GMinucci: Having a movie list is a pre-requisite to the method execution so we should use a `guard` statement.
```suggestion
            guard 
                let self = self,
                let movieList = movieList else
            { return }
            self.movies = movieList
```

ViewControllers/MostPopularCollectionViewController.swift
```diff
@@ -91,16 +82,34 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
...
+            
+            DispatchQueue.main.async { [weak self] in
+                self?.collectionView.reloadData()
+                self?.itemInViewIndex = Int(round(Double((self?.movies.count)! / 2)))
+                self?.collectionView.scrollToItem(at: IndexPath(item: self!.itemInViewIndex, section: 0), at: UICollectionView.ScrollPosition.centeredHorizontally, animated: false)
+                self?.configureWithData(index: self!.itemInViewIndex)
+                self?.fadeIn()
+                }
```
GMinucci: Since we are safely checking for `self` existence in the lines above we don't need to keep the optional self here. Also, this block is not going to retain self anymore so we don't need this second `[weak self]`
```suggestion
            DispatchQueue.main.async {
                self.collectionView.reloadData()
                self.itemInViewIndex = Int(round(Double(self.movies.count / 2)))
                self.collectionView.scrollToItem(at: IndexPath(item: self.itemInViewIndex, section: 0), at: UICollectionView.ScrollPosition.centeredHorizontally, animated: false)
                self.configureWithData(index: self.itemInViewIndex)
                self.fadeIn()
                }
```

ViewControllers/MostPopularCollectionViewController.swift
```diff
@@ -91,16 +82,34 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
...
+        setupUI()
+        
+        movieDataService.getMostPopularMoviesList { [weak self] (moviesList: [Movie]?) in
+            if let moviesList = moviesList {
+                self?.movies = moviesList
+            }
+            
+            DispatchQueue.main.async { [weak self] in
+                self?.collectionView.reloadData()
+                self?.itemInViewIndex = Int(round(Double((self?.movies.count)! / 2)))
```
GMinucci: Quick note: Careful with force casting properties `!`. Unless you are sure that it will never be `nil` this can easily crash your app, I think that if you open movie details and close it super fast (before the API request is finished, the app will probably crash at the moment). We can discuss a bit more about weak/strong properties if you want 😁

ViewControllers/MostPopularCollectionViewController.swift
```diff
@@ -91,16 +82,34 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
...
+                self?.collectionView.scrollToItem(at: IndexPath(item: self!.itemInViewIndex, section: 0), at: UICollectionView.ScrollPosition.centeredHorizontally, animated: false)
```
GMinucci: Same as the other comment about force casting

ViewControllers/MostPopularCollectionViewController.swift
```diff
@@ -192,7 +187,7 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
 //        }
     }
 
-    func configureWithData(index: Int) {
+    private func configureWithData(index: Int) {
         let movieName = movies[index].title
         let reviewsScore = "\(movies[index].voteAverage)%"
         let movieDescription = movies[index].overview
```
GMinucci: Just a quick tip, instead of accessing the array at index on every line you can do it in a single property and reuse it so you avoid miss typing the index. Also, improves readability
```suggestion
        let movie = movies[index]
        let movieName = movie.title
        let reviewsScore = "\(movie.voteAverage)%"
        let movieDescription = movie.overview
```

ViewControllers/PlayingNowCollectionViewController.swift
```diff
@@ -53,26 +56,17 @@ class PlayingNowCollectionViewController: UICollectionViewController, UICollecti
...
-        let reviewsScore = "\(movies[indexPath.item].voteAverage)%"
-        let path = movies[indexPath.item].posterPath
-        if let url = movieDataService.getMoviePosterURL(posterPath: path) {
-            cell.configure(imageURL: url, name: movieName, duration: "0h0", reviewsScore: reviewsScore)
+        let movie = movies[indexPath.item]
+        let movieName = movie.title
+        let reviewsScore = movie.voteAverage
+        let popularity = movie.popularity
+        if let path = movie.posterPath {
+            cell.configure(imageURL: path, name: movieName, reviewsScore: reviewsScore, popularity: popularity)
```
GMinucci: Let's make the `cell.configure` method accept an optional imageURL, so we will always call it even when there is no poster path. The way we have here, in case a move has no poster image it will show as a completely empty cell even though we have other data such name, popularity, etc.

Views/DetailCollectionViewCell.swift
```diff
@@ -0,0 +1,60 @@
...
+//  Movies
+//
+//  Created by Diana Chizhik on 23/06/2022.
+//
+
+import Foundation
+import UIKit
+
+class DetailCollectionViewCell: UICollectionViewCell {
+    static var reuseIdentifier: String { NSStringFromClass(self) }
```
GMinucci, line 12: 👏

Views/DetailCollectionViewCell.swift
```diff
@@ -0,0 +1,60 @@
...
+        NSLayoutConstraint.activate([
+            label.topAnchor.constraint(equalTo: contentView.topAnchor, constant: 4),
+            label.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 15),
+            label.trailingAnchor.constraint(equalTo: contentView.trailingAnchor, constant: -15),
+            label.bottomAnchor.constraint(equalTo: contentView.bottomAnchor, constant: -4),
+//
+//            contentView.leadingAnchor.constraint(equalTo: leadingAnchor),
+//            contentView.topAnchor.constraint(equalTo: topAnchor),
+//            contentView.trailingAnchor.constraint(equalTo: trailingAnchor),
+//            contentView.bottomAnchor.constraint(equalTo: bottomAnchor)
```
GMinucci: Let's burn old comments. In git we trust 🙏

GMinucci (changes requested):

ViewControllers/MostPopularCollectionViewController.swift
DiChizhik: I do lack consistency, since I did it in some places and didn't in others

Application/SceneDelegate.swift
DiChizhik: I believe I left this comment since I have a question re that :)

Models/Movie.swift
DiChizhik: So, it's going to be

if let posterPathString = try values.decodeIfPresent(String.self, forKey: .posterPath) {
            posterPath = URL(string: "https://image.tmdb.org/t/p/w500" + posterPathString)
        } else {
            posterPath = nil
        }

ViewControllers/MostPopularCollectionViewController.swift
DiChizhik: I'm not sure I understand which comment you're referring to.

GMinucci: This one here: https://github.com/DiChizhik/Movies/pull/1#discussion_r905900218

Resources/Assets.xcassets/mostRecentIcon.imageset/Contents.json
DiChizhik: I guess you've said it before, so I downloaded only .svg

ViewControllers/DetailViewController.swift
DiChizhik: I thought it might come in handy when refactoring the ugly way to display genres and languages.
These properties help to test the implementation.
I'll leave them for now.

Services/MovieDataService.swift
```diff
@@ -9,58 +9,66 @@ import Foundation
 import UIKit
 
 protocol MovieDataServiceProtocol {
-    func getPlayingNowMoviesList(completion: @escaping ([Movie]?)-> Void)
-    func getMostPopularMoviesList(completion: @escaping ([Movie]?)-> Void)
+    func getPlayingNowMoviesList(completion: @escaping (Result<[Movie], MovieServiceError>)-> Void)
+    func getMostPopularMoviesList(completion: @escaping (Result<[Movie], MovieServiceError>)-> Void)
+}
+
+enum MovieServiceError: Error {
```
GMinucci, line 16: 👏

Services/MovieDataService.swift
```diff
@@ -9,58 +9,66 @@ import Foundation
...
-    func getPlayingNowMoviesList(completion: @escaping ([Movie]?)-> Void) {
+    func getPlayingNowMoviesList(completion: @escaping (Result<[Movie], MovieServiceError>)-> Void) {
         guard let url = URL(string: "https://api.themoviedb.org/3/movie/popular?api_key=b2b14caf40262a9c19a366b15e4e3537&language=en-US&page=1") else { return }
 
         let request = URLRequest(url: url)
         URLSession.shared.dataTask(with: request) {(data, response, error) in
             if let error = error {
                 print(error.localizedDescription)
-                completion(nil)
+//                How can I use error.localizedDescription to display the error in the VC?
```
GMinucci, line 31:
```swift
// First you conform to `LocalizedError` protocol
enum MovieServiceError: Error, LocalizedError {
    case failedToDecode, noInternet
    
    // Then you add this computed property. This will be the message returned when you call `error.localizedDescription`
    var errorDescription: String? {
        switch self {
        case .failedToDecode:
            return "Failed to decode API response"
            
        case .noInternet:
            return "No internet connection"
        }
    }
}
```

ViewControllers/DetailViewController.swift
```diff
@@ -147,14 +145,28 @@ class DetailViewController: UIViewController {
...
-                self?.movieDetails = movieDetails
+        
+        movieDataService.getMovieDetails(movieId: id) { [weak self] result in
+            guard let self = self else {return}
+            
+            switch result {
+            case .success(let details):
+                self.movieDetails = details
+            case .failure(let error):
+//                Here I wanted to display an alert controller with the info about the error. Not sure if error.localizedDescription will hold anything though.
```
GMinucci, line 156: At the moment no since your error doesn't conform to `LocalizedError`. I've left a comment on how to do that :)

ViewControllers/LaunchViewController.swift
```diff
@@ -0,0 +1,97 @@
...
+                       completion: { done in
+                            if done {
+                                UIView.animate(withDuration: 1,
+                                               animations: {
+                                                    self.logoImageView.alpha = 0.5
+                                                    self.logoImageView.transform = CGAffineTransform(scaleX: 83, y: 83)
+                                                },
+                                               completion: { done in
+                                                    if done {
+                                                        DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
```
GMinucci, line 53: Nice! UI updates always on main thread 🙌

GMinucci (approved):
Looking great!

PR #2, opened 2022-06-29
Pagination, collectionView, scrollView

DiChizhik:


Services/MovieDataService.swift
```diff
@@ -13,61 +13,87 @@ protocol MovieDataServiceProtocol {
...
+            return "Failed to load data"
+        }
+    }
 }
 
 class MovieDataService: MovieDataServiceProtocol {
+    var playingNowPage = 1
+    var mostPopularPage = 1
+    var isPlayingNowRequestCompleted = true
+    var isMostPopularRequestCompleted = true
```
GMinucci, line 37: Let's make it private. We don't want other classes messing around with these properties 😁
```suggestion
    private var playingNowPage = 1
    private var mostPopularPage = 1
    private var isPlayingNowRequestCompleted = true
    private var isMostPopularRequestCompleted = true
```

Services/MovieDataService.swift
```diff
@@ -13,61 +13,87 @@ protocol MovieDataServiceProtocol {
...
                 return
             }
-            
+                   
             guard let data = data else {
                 completion(.failure(.failedToGetData))
                 return
             }
+             
+            self.isPlayingNowRequestCompleted = true
```
GMinucci: Let's move this to be called as soon as the `dataTask` finishes, on line 46. In the scenario here, in case you get error it will return without setting `isPlayingNowRequestCompleted` back to true making it impossible to fetch more data afterwards.

Services/MovieDataService.swift
```diff
@@ -13,61 +13,87 @@ protocol MovieDataServiceProtocol {
...
             
             let jsonDecoder = JSONDecoder()
             guard let moviesData = try? jsonDecoder.decode(Movies.self, from: data) else {
-                completion(.failure(.failedToDecodeData))
+                completion(.failure(.failedToDecode))
```
GMinucci, line 61: Great job on the error parsing! 👏

Services/MovieDataService.swift
```diff
@@ -13,61 +13,87 @@ protocol MovieDataServiceProtocol {
...
                 return
             }
-            
+                    
             guard let data = data else {
                 completion(.failure(.failedToGetData))
                 return
             }
             
+            self.isMostPopularRequestCompleted = true
```
GMinucci: Same as previously, we need to set `isMostPopularRequestCompleted` back to true as soon as the data task is done, regardless of success or not

ViewControllers/DetailViewController.swift
```diff
@@ -141,36 +152,42 @@ class DetailViewController: UIViewController {
...
-                    ac.addAction(UIAlertAction(title: "OK", style: .default))
-                    self.present(ac, animated: true)
-                default:
-                    break
+                case .failedToGetResponse:
+                    self.showError(message: error.localizedDescription)
+                case .failedToDecode:
+                    self.showError(message: error.localizedDescription)
+                case .failedToGetData:
+                    self.showError(message: error.localizedDescription)
```
GMinucci: Since we are not doing anything different between errors this can be simply:
```swift
self.showError(message: error.localizedDescription)
```
Without the `switch` between error cases

ViewControllers/DetailViewController.swift
```diff
@@ -141,36 +152,42 @@ class DetailViewController: UIViewController {
...
-        
+            
             DispatchQueue.main.async {
                 self.configureWithData()
                 self.setupUI()
             }
         }
     }
     
+    private func showError(message: String) {
```
GMinucci: Nice! This is a great step towards standardizing error alerts through the app, to have a consistent UX 😉

ViewControllers/DetailViewController.swift
```diff
@@ -186,8 +203,27 @@ class DetailViewController: UIViewController {
...
+            scrollView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            scrollView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
+            scrollView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor),
+            
+            contentView.leadingAnchor.constraint(equalTo: scrollView.contentLayoutGuide.leadingAnchor),
+            contentView.topAnchor.constraint(equalTo: scrollView.contentLayoutGuide.topAnchor),
+            contentView.trailingAnchor.constraint(equalTo: scrollView.contentLayoutGuide.trailingAnchor),
+            contentView.bottomAnchor.constraint(equalTo: scrollView.contentLayoutGuide.bottomAnchor),
+            
+            contentView.widthAnchor.constraint(equalTo: view.widthAnchor)
```
GMinucci: 🔥

ViewControllers/DetailViewController.swift
```diff
@@ -111,7 +123,6 @@ class DetailViewController: UIViewController {
         collectionView.register(DetailCollectionViewCell.self, forCellWithReuseIdentifier: DetailCollectionViewCell.reuseIdentifier)
         collectionView.dataSource = self
         collectionView.delegate = self
-        collectionView.isScrollEnabled = false
```
GMinucci, line 114: Carefull when adding a scrollview inside a scrollview (in this case we have the scrollview and a collection inside which is also a scrollview). This can lead to some weird funky behaviours. To make things easier for us we usually lock all scrollviews and handle scrolling on one of them.

ViewControllers/PlayingNowCollectionViewController.swift
```diff
@@ -20,6 +20,7 @@ class PlayingNowCollectionViewController: UICollectionViewController {
         
         collectionView.register(PlayingNowCollectionViewCell.self, forCellWithReuseIdentifier: PlayingNowCollectionViewCell.reuseIdentifier)
 
+        // Shall I make it a separate method to reuse it in willDisplayCell method?
```
GMinucci: Seems a good idea. Improves readability and you can reuse it anywhere that is needed

Views/SelfSizingCollectionView.swift
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        commonInit()
+    }
+
+    private func commonInit() {
+        isScrollEnabled = false
+    }
```
GMinucci: Uh I see why you removed the `isScrollEnabled` from the class! Good one.
But you are not using this class there 🤔

ViewControllers/DetailViewController.swift
```diff
@@ -202,40 +238,67 @@ class DetailViewController: UIViewController {
...
             movieDescriptionLabel.leftAnchor.constraint(equalTo: releaseStack.leftAnchor),
-            movieDescriptionLabel.rightAnchor.constraint(equalTo: view.layoutMarginsGuide.rightAnchor),
+            movieDescriptionLabel.rightAnchor.constraint(equalTo: contentView.layoutMarginsGuide.rightAnchor),
             movieDescriptionLabel.topAnchor.constraint(equalTo: durationStack.bottomAnchor, constant: 16),
 
             collectionView.leftAnchor.constraint(equalTo: releaseStack.leftAnchor),
             collectionView.topAnchor.constraint(equalTo: movieDescriptionLabel.bottomAnchor, constant: 16),
             collectionView.rightAnchor.constraint(equalTo: movieDescriptionLabel.rightAnchor),
-            collectionView.heightAnchor.constraint(equalToConstant: 200),
+            collectionView.heightAnchor.constraint(equalToConstant: calculateCollectionViewHeight())
```
GMinucci: Things to keep in mind here:
- Even though there is a method to calculate collection height, this is called only once and the collection height constraint will be static. In our view here this is working because `setupUI` is called only after data is retrieved so the collection can be sized properly. Ideally `setupUI` should be called before the view is even displayed (at `loadView` method for example). In some cases our screen will have a "loading state" to be shown before content is retrieved.
- There is no constraint from collection to the bottom of the content view. For the content view to be able to know its own size, everything should be linked internally. In our case here, all subviews are properly interconnected up until the last link, from bottom of collection to bottom of content view. If we add that connection then your content view will be able to know its own size dynamically depending on its subviews sizes.

In case you want/need to have a constraint that you can update manually you can do like this:
```swift
class MyViewController: UIViewController {
    private var myHeightConstraint: NSLayoutConstraint?
    private let myView = UIView()
    
    override func viewDidLoad() {
        super.viewDidLoad()
        
        view.addSubview(myView)
        myHeightConstraint = myView.heightAnchor.constraint(equalToConstant: 10)
        myHeightConstraint.isActive = true
    }

    private func updateHeight(to newHeight: CGFloat) {
        myHeightConstraint.constant = newHeight
    }
}
```

ViewControllers/DetailViewController.swift
```diff
@@ -202,40 +238,67 @@ class DetailViewController: UIViewController {
...
+    
+    private func calculateScrollViewContentSize()-> CGFloat {
+        let lastItemOrigin = collectionView.frame.origin.y
+        let lastItemHeight = collectionView.frame.size.height
+        
+        print(lastItemHeight)
+        
+        let scrollViewContentSize = lastItemHeight + lastItemOrigin
+        return scrollViewContentSize
+    }
```
GMinucci: It's fine manually calculating the collection content size, but we can use UIKit itself to tell us the size and leave the work of calculating it to autolayout itself.
```swift
collectionView.setNeedsLayout() // invalidates current layout and triggers a layout update in the next update cycle
collectionView.layoutIfNeeded() // updates the layout immediately if needed
let updatedContentSize = collectionView.contentSize
```

ViewControllers/DetailViewController.swift
```diff
@@ -202,40 +238,67 @@ class DetailViewController: UIViewController {
...
             collectionView.topAnchor.constraint(equalTo: movieDescriptionLabel.bottomAnchor, constant: 16),
             collectionView.rightAnchor.constraint(equalTo: movieDescriptionLabel.rightAnchor),
-            collectionView.heightAnchor.constraint(equalToConstant: 200),
+            collectionView.heightAnchor.constraint(equalToConstant: calculateCollectionViewHeight())
         ])
     }
     
+    override func viewDidLayoutSubviews() {
+        super.viewDidLayoutSubviews()
+        scrollView.contentSize = CGSize(width: view.frame.size.width, height: calculateScrollViewContentSize())
```
GMinucci: We are only needing to do this here because our `contentView` doesn't know its size. We are missing the link between the bottom of the last view and the bottom of our content view. After that is set, this work will be automatically done by the layout engine itself

ViewControllers/DetailViewController.swift
```diff
@@ -202,40 +238,67 @@ class DetailViewController: UIViewController {
...
+    private func calculateCollectionViewHeight()-> CGFloat {
+        var languageRows = languages.count / 4
+        languages.count % 4 != 0 ? languageRows += 1 : nil
+        let languageRowsHeight = (languageRows * 27) + ((languageRows - 1) * 4)
+        
+        var genreRows = genres.count / 4
+        genres.count % 4 != 0 ? genreRows += 1 : nil
+        let genreRowsHeight = (genreRows * 27) + ((genreRows - 1) * 4)
+        
+        return CGFloat(languageRowsHeight + 16 + genreRowsHeight)
```
GMinucci: If we choose to calculate the collection height manually we should be using the content that is displayed in the collection instead of languages and rows lists. Yes, they are being displayed here, but it can happen that it's not in sync with what is actually being displayed which will lead to wrong sizes. We can use the collection datasource methods here to get number of sections and items:

```swift
var totalHeight: CGFloat = 0
let numberOfSections = collectionView.numberOfSections
for section in 0..<numberOfSections {
    let numberOfItems = collectionView.numberOfItems(inSection: section)
    // calculate section height here and add to "totalHeight"
}
return totalHeight
```
Additionally, this code is generic so you can reuse it in any collection view regardless of the content

GMinucci (changes requested):

Views/SelfSizingCollectionView.swift
DiChizhik: I found it, I didn't fully understand it, so I left it out to ask about later :)

DiChizhik: and forgot about disabling scrollView again🤦

ViewControllers/DetailViewController.swift
DiChizhik: Can we make it universal like services?
So that other viewControllers have access to it?

ViewControllers/DetailViewController.swift
DiChizhik: I tried pinning the collectionView to the bottom of the contentView but it just failed to display, all sizes being zero

ViewControllers/DetailViewController.swift
GMinucci: Sounds like a really good idea. You can either make it an extension of UIViewController or maybe a service to handle it. Both ways have its own pros and cons.

ViewControllers/MostPopularViewController.swift
```diff
@@ -0,0 +1,194 @@
...
+    @objc func seeMoreTapped(_ sender: UIButton) {
+        let selectedMovieId = movies[itemInViewIndex].id
+        let detailViewController = MovieDetailViewController(selectedMovieID: selectedMovieId)
+        let detailNavigationController = UINavigationController(rootViewController: detailViewController)
+        present(detailNavigationController, animated: true)
+    }
+}
+
+// I've put all scroll-related methods here.
+extension MostPopularViewController {
```
GMinucci: In case you want to move all code related to delegate implementation to a separate extension, the scroll methods you see here are coming from a `UIScrollViewDelegate`. `UICollectionViewDelegate` inherits from `UIScrollViewDelegate`, so that's why you have then when you conform to it. In your case here the conformance is coming from the class declaration at line 11.
Not actually a change request, more of an information :)

ViewControllers/MostPopularViewController.swift
```diff
@@ -0,0 +1,194 @@
...
+        guard let cell = collectionView.dequeueReusableCell(withReuseIdentifier: MostPopularCollectionViewCell.reuseIndentifier, for: indexPath) as? MostPopularCollectionViewCell else { fatalError() }
+            
+        let path = movies[indexPath.item].posterPath
+        cell.configure(imageURL: path)
+        
+        return cell
+    }
+}
+
+// Why is conformance to UICollectionViewDelegate here redundant?
```
GMinucci: Not sure if I got the question. You mean conforming to `UICollectionViewDelegate` at the top and then `UICollectionViewDelegateFlowLayout` here? If that's the question then yes, you don't need to conform twice, just conforming to `UICollectionViewDelegateFlowLayout` is enough as it inherits from the `UICollectionViewDelegate` anyways. You can leave both conformance as well for readability or any other reason you want, but technically not necessary.

ViewControllers/MovieDetailViewController.swift
```diff
@@ -0,0 +1,153 @@
...
+//  Movies
+//
+//  Created by Diana Chizhik on 07/06/2022.
+//
+
+import UIKit
+import Kingfisher
+
+class MovieDetailViewController: UIViewController {
+    static var identifier: String{NSStringFromClass(self)}
```
GMinucci: I don't think this is being used anymore. We are using the identifiers from the cell itself now.

ViewControllers/MovieDetailViewController.swift
```diff
@@ -0,0 +1,153 @@
...
+        return view
+    }()
+
+    init(selectedMovieID: Int) {
+        self.selectedMovieID = selectedMovieID
+        
+        super.init(nibName: nil, bundle: nil)
+    }
+    
+//    what does this line mean?
```
GMinucci: Since we added a custom init to this class, swift requires us to write all initializers that are marked as `required`. So we are required to write the `init?(coder: NSCoder)` method. This init is only called when this class is initialized from a coder, for example if this class was being called from a storyboard.

Additionally we have a decorator in the method:
`@available(*, unavailable)`
Decorator is a design pattern that allows you to add a behaviour to a method/object/property without changing it. In this case, `@available` adds an "availability" behaviour that tells Xcode whether this method is available or not. First parameter we specify the OS (* means any OS) and second parameter we say that is unavailable. So in case someone tries to call this method in your code, Xcode will not run and show an error at that line.

Views/LeftAlignedCollectionViewFlowLayout.swift
```diff
@@ -0,0 +1,32 @@
...
+//  LeftAlignedCollectionViewFlowLayout.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 01/07/2022.
+//
+
+import Foundation
+import UIKit
+
+class LeftAlighedCollectionViewFlowLayout: UICollectionViewFlowLayout {
```
GMinucci: Small typo
```suggestion
class LeftAlignedCollectionViewFlowLayout: UICollectionViewFlowLayout {
```

Views/LeftAlignedCollectionViewFlowLayout.swift
```diff
@@ -0,0 +1,32 @@
+//
+//  LeftAlignedCollectionViewFlowLayout.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 01/07/2022.
+//
+
+import Foundation
+import UIKit
```
GMinucci: If we import `UIKit` we don't need to import `Foundation`. Internally, UIKit depends on Foundation anyways

Views/MostPopularView.swift
```diff
@@ -0,0 +1,134 @@
...
+        description.font = UIFont.systemFont(ofSize: 15)
+        description.textColor = UIColor(named: "descriptionColor")
+        description.textAlignment = .center
+        description.alpha = 0
+        return description
+    }()
+    
+    private(set) lazy var seeMoreButton: UIButton = {
+        let button = UIButton(type: .system)
+//        button.addTarget(self, action: #selector(seeMoreTapped), for: .touchUpInside)
```
GMinucci: We can delete this comment.
Also, there are 2 ways to achieve this. One is simply adding a target to the button from the Controller class directly (as you did, and it's usually the common way) or you can define a delegate for the view and make the Controller conform to this delegate protocol, then you detect the tap in the View but pass the responsability to the Controller to handle it.
This is the ideal implementation in a perfect world, but yeah... I don't have any problems with the 1st way and I feel it's good enough (less boilerplate code, easier to read), only downside is that is harder to test. So there are pros and cons 😉

Views/StackMovieDetailView.swift
```diff
@@ -0,0 +1,177 @@
...
+        description.textAlignment = .left
+        return description
+    }()
+    
+    private(set) lazy var collectionViewFlowLayout: UICollectionViewFlowLayout = {
+        let layout = LeftAlighedCollectionViewFlowLayout()
+        layout.minimumInteritemSpacing = 4
+        layout.minimumLineSpacing = 4
+        layout.sectionInset = UIEdgeInsets(top: 4, left: 0, bottom: 0, right: 0)
+//       adjusted
```
GMinucci: 👀

Views/StackMovieDetailView.swift
```diff
@@ -0,0 +1,177 @@
...
+        let scroll = UIScrollView()
+        scroll.translatesAutoresizingMaskIntoConstraints = false
+        return scroll
+    }()
+    
+    private(set) lazy var scrollContentStack:UIStackView = {
+        let stackView = UIStackView()
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.axis = .vertical
+//        consider this
```
GMinucci, line 109: 👀

Views/StackMovieDetailView.swift
```diff
@@ -0,0 +1,177 @@
...
+        super.init(frame: frame)
+        setupUI()
+    }
+    
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        setupUI()
+    }
+    
+    private func setupUI() {
```
GMinucci, line 125: So you liked the stack view approach more 😎

GMinucci (approved):
Great work! Just added some small comments and notes

ViewControllers/MostPopularViewController.swift
DiChizhik: Right, conformance to UICollectionViewDelegate has been stated above. I failed to notice.
Xcode complained when I wanted to add it to the extension.

ViewControllers/MovieDetailViewController.swift
DiChizhik: Do you mean it's redundant?
I don't even remember why I created it. must have been at the very beginning

ViewControllers/MovieDetailViewController.swift
DiChizhik: Right, my question was about the decorator.
So, we have it here since the implementation of the required init? is necessary but we don't actually need it?

Views/StackMovieDetailView.swift
DiChizhik: I've used it less, so wanted to try it. I plan to try creating a stackView in mostPopularView, too.

ViewControllers/MovieDetailViewController.swift
GMinucci: The init itself is required, the decorator is not required indeed. It's just a way to make the code safer and future proof (imagine someone joins the project and tries to call this class from this init, with this he will know that he is not supposed to do that).

ViewControllers/MostPopularViewController.swift
```diff
@@ -36,11 +41,11 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
         
         loadMovieData()
         
-//        I did it in an attemt to make loadMovieData method reusable in willDisplayCell method. I'm not sure it's right, though.
+//      I did it in an attemt to make loadMovieData method reusable in willDisplayCell method. I'm not sure it's right, though.
         DispatchQueue.main.async {
             self.contentView.collectionView.scrollToItem(at: IndexPath(item: self.itemInViewIndex, section: 0), at: UICollectionView.ScrollPosition.centeredHorizontally, animated: false)
             self.configureWithData(index: self.itemInViewIndex)
-            self.fadeIn()
+            self.fade(Fading.fadeIn)
```
GMinucci: Usually with enums we can take advantage of swift's type inference and write like this:
```swift
self.fade(.fadeIn)
```
Not a change request here, just a FYI

ViewControllers/MostPopularViewController.swift
```diff
@@ -86,68 +91,28 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
         }
     }
     
-    @objc func seeMoreTapped(_ sender: UIButton) {
+    @objc internal func seeMoreTapped() {
```
GMinucci: I'm curious why we went for a `internal` here, we can discuss in the meeting Today

ViewControllers/MostPopularViewController.swift
```diff
@@ -17,10 +17,15 @@ class MostPopularViewController: UIViewController, UICollectionViewDelegate {
...
         view.collectionView.delegate = self
-        view.seeMoreButton.addTarget(self, action: #selector(seeMoreTapped), for: .touchUpInside)
+        view.delegate = self
         return view
     }()
     
+    private enum Fading: Int {
+        case fadeIn = 1
+        case fadeOut = 0
+    }
```
GMinucci: Nicely done!

Views/MostPopularView.swift
```diff
@@ -5,15 +5,20 @@
...
+}
+
 class MostPopularView: UIView {
     private(set) lazy var collectionViewFlowLayout: UICollectionViewFlowLayout = {
         let layout = UICollectionViewFlowLayout()
         layout.scrollDirection = .horizontal
         layout.minimumInteritemSpacing = 32
         layout.minimumLineSpacing = 30
+// Can I somehow define it here or doing that in separate methods in the viewController is the only way?
+
```
GMinucci: You should be able to set it here, is it not possible?

Views/MostPopularView.swift
```diff
@@ -5,15 +5,20 @@
 //  Created by Diana Chizhik on 04/07/2022.
 //
 
-import Foundation
 import UIKit
 
+protocol MostPopularViewDelegate: AnyObject {
+    func seeMoreTapped()
+}
```
GMinucci, line 12: 🔥

GMinucci (approved):

PR #3, opened 2022-06-29
Enable scrolling in Movie details screen

GMinucci:
This PR will enable scrolling in the movie details view. 

I've added 2 variants of the view implementation. One using only NSLayoutConstraints and other using UIStackView. Feel free to choose which one you like more 😁

Notes:

I've moved the view related code to a dedicated file, this way we will be following MVC properly. If you want more details on the architecture you can check [here](https://www.raywenderlich.com/1000705-model-view-controller-mvc-in-ios-a-modern-approach). Also, if you want we can make a call and I can talk a bit more about architecture (what it is, common ones, etc). This change is purely to organize the code.

Lastly, I did a small refactor in the `DetailsViewController`, just removed unused code and moved some methods around. If you don't agree on any of the changes feel free to comment in the PR.

### DIFF
--- a/Utils/IntrinsicallySizedCollectionView.swift
+++ b/Utils/IntrinsicallySizedCollectionView.swift
@@ -1,0 +1,28 @@
+//
+//  IntrinsicallySizedCollectionView.swift
+//  Movies
+//
+//  Created by Gabriel Rodrigues Minucci on 29/06/2022.
+//
+
+import UIKit
+
+class IntrinsicallySizedCollectionView: UICollectionView {
+    
+    override var contentSize: CGSize {
+        didSet {
+            invalidateIntrinsicContentSize()
+        }
+    }
+    
+    override func reloadData() {
+        super.reloadData()
+        invalidateIntrinsicContentSize()
+    }
+    
+    override var intrinsicContentSize: CGSize {
+        setNeedsLayout()
+        layoutIfNeeded()
+        return contentSize
+    }
+}
--- a/Utils/LeftAlignedCollectionViewFlowLayout.swift
+++ b/Utils/LeftAlignedCollectionViewFlowLayout.swift
@@ -1,0 +1,29 @@
+//
+//  LeftAlignedCollectionViewFlowLayout.swift
+//  Movies
+//
+//  Created by Gabriel Rodrigues Minucci on 29/06/2022.
+//
+
+import UIKit
+
+class LeftAlignedCollectionViewLayout: UICollectionViewFlowLayout {
+    override func layoutAttributesForElements(in rect: CGRect) -> [UICollectionViewLayoutAttributes]? {
+        let attributes = super.layoutAttributesForElements(in: rect)
+
+        var leftMargin = sectionInset.left
+        var maxY: CGFloat = -1.0
+        attributes?.forEach { layoutAttribute in
+            if layoutAttribute.frame.origin.y >= maxY {
+                leftMargin = sectionInset.left
+            }
+
+            layoutAttribute.frame.origin.x = leftMargin
+
+            leftMargin += layoutAttribute.frame.width + minimumInteritemSpacing
+            maxY = max(layoutAttribute.frame.maxY , maxY)
+        }
+
+        return attributes
+    }
+}
--- a/Utils/LeftAlignedCollectionViewFlowLayout.swift
+++ b/Utils/LeftAlignedCollectionViewFlowLayout.swift
@@ -9,18 +9,33 @@ import UIKit
 
 class LeftAlignedCollectionViewLayout: UICollectionViewFlowLayout {
     override func layoutAttributesForElements(in rect: CGRect) -> [UICollectionViewLayoutAttributes]? {
+        // atributes is a list of layout attributes for the elements that are currently being displayed in our fram
+        // we firstly get the default attributes by calling the super method
         let attributes = super.layoutAttributesForElements(in: rect)
 
+        // our left margin starts at the left most side (0) + our sectionInset.left
         var leftMargin = sectionInset.left
+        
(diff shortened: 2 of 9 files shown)


Utils/IntrinsicallySizedCollectionView.swift
```diff
@@ -0,0 +1,28 @@
...
+    override func reloadData() {
+        super.reloadData()
+        invalidateIntrinsicContentSize()
+    }
+    
+    override var intrinsicContentSize: CGSize {
+        setNeedsLayout()
+        layoutIfNeeded()
+        return contentSize
+    }
```
DiChizhik, line 27: I've seen this solution as I was searching for an answer, but I didn't understand it. Would you explain it to me during our next call, please?

Utils/LeftAlignedCollectionViewFlowLayout.swift
```diff
@@ -0,0 +1,29 @@
...
+
+            layoutAttribute.frame.origin.x = leftMargin
+
+            leftMargin += layoutAttribute.frame.width + minimumInteritemSpacing
+            maxY = max(layoutAttribute.frame.maxY , maxY)
+        }
+
+        return attributes
+    }
+}
```
DiChizhik, line 44: I'm afraid, you've lost me here:

maxY = max(layoutAttribute.frame.maxY , maxY)

 I could use an explanation too🙏

ViewControllers/MovieDetail/ConstraintMovieDetailView.swift
```diff
@@ -0,0 +1,201 @@
...
+    private(set) lazy var titleLabel: UILabel = {
+        let label = UILabel()
+        label.translatesAutoresizingMaskIntoConstraints = false
+        label.font = UIFont.systemFont(ofSize: 20, weight: .heavy)
+        label.textAlignment = .center
+        label.textColor = UIColor.white
+        label.numberOfLines = 0
+        return label
+    }()
+
```
DiChizhik, line 20: Why have you added (set) to the access modifier? Why is it important here?

ViewControllers/MovieDetail/StackMovieDetailView.swift
```diff
@@ -0,0 +1,171 @@
...
+        scrollView.translatesAutoresizingMaskIntoConstraints = false
+        return scrollView
+    }()
+    
+    private(set) lazy var scrollContentStack: UIStackView = {
+        let stack = UIStackView()
+        stack.translatesAutoresizingMaskIntoConstraints = false
+        stack.axis = .vertical
+        stack.layoutMargins = UIEdgeInsets(top: 0, left: 16, bottom: 0, right: 16)
+        stack.isLayoutMarginsRelativeArrangement = true
```
DiChizhik, line 102: If you hadn't set this property to 'true', would all views be left-aligned? or what would happen?

Utils/IntrinsicallySizedCollectionView.swift
GMinucci: For sure :) 
Please remind me in case I forget

Utils/LeftAlignedCollectionViewFlowLayout.swift
GMinucci: I've added comments on this class to explain what is happening there. Feel free to ask more about it in our next session if you still have questions.
In case the question was just about the `max` method, it's a math method that returns the max value between parameters, for example:
```swift
let a = max(3, 4) // a is 4
let a = min(3, 4) // a is 3
```

ViewControllers/MovieDetail/ConstraintMovieDetailView.swift
GMinucci: It's not really necessary, but I like limiting access from other classes and try to only expose what is really necessary. Since we need to expose the view elements here (labels, collection, imageview, etc) I can not leave them as `private`, so be able to only get them I can set `private(set)` which means that setting is private but getting is public.
```swift
class MyClass {
    private(set) label = UILabel()
}

let a  = MyClass()
a.label.text = "text" // this is allowed
a.label = UILabe() // this won't be allowed
```

ViewControllers/MovieDetail/StackMovieDetailView.swift
GMinucci: I need to set `isLayoutMarginsRelativeArrangement` to the stack view to propagate the `layoutMargins` that I set above it. By default the stack will layout its arranged subviews according to its own bounds (entire view), if I set it to true then it layout its arranged subviews following the layout margins
Stack view alignment goes to `stackView.alignment` property. Defaults to `.fill` but you can set to `.leading` to be "left" aligned depending on your device region

Utils/LeftAlignedCollectionViewFlowLayout.swift
DiChizhik: I just didn't get the last line. Now I do! Amazing, love it!

ViewControllers/MovieDetail/ConstraintMovieDetailView.swift
DiChizhik: I see, thanks!

ViewControllers/MovieDetail/StackMovieDetailView.swift
DiChizhik: Got it!
That was a stupid question, really

Utils/IntrinsicallySizedCollectionView.swift
```diff
@@ -0,0 +1,28 @@
...
+}
```
DiChizhik, line 28: This class is something that would be great to discuss on Tue!

GMinucci:
This feature was already accomplished after merging #2.

PR #4, opened 2022-07-06
Reusable protocol

GMinucci:
This PR adds the Reusable protocol. This protocol adds a functionality to any target class to have its own reuse identifier automatically.
This also adds some convenience methods to handle UITableViewCell and UICollectionViewCell register and dequeue operations.

### DIFF
--- a/Protocols/Reusable.swift
+++ b/Protocols/Reusable.swift
@@ -1,0 +1,60 @@
+//
+//  Reusable.swift
+//  Movies
+//
+//  Created by Gabriel Rodrigues Minucci on 06/07/2022.
+//
+
+import UIKit
+
+protocol Reusable: AnyObject {
+    static var reuseIdentifier: String { get }
+}
+
+extension Reusable {
+    // https://forums.swift.org/t/relying-on-string-describing-to-get-the-name-of-a-type/16391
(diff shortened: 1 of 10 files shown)


Protocols/Reusable.swift
```diff
@@ -0,0 +1,60 @@
...
+
+extension Reusable {
+    // https://forums.swift.org/t/relying-on-string-describing-to-get-the-name-of-a-type/16391
+    static var reuseIdentifier: String { NSStringFromClass(self) }
+}
+
+// MARK: - Table View convenience methods
+
+extension UITableView {
+    func register<Cell>(_ cellType: Cell.Type) where Cell: UITableViewCell, Cell: Reusable {
```
DiChizhik, line 22: So, it a generic function, right?
Are you using it here because of the protocol, to ensure that UITableView that calls this function takes only Reusable-conforming UITableViewCells?

GMinucci: Exactly 😁
It can also be written as:
```swift
func register<Cell: UITableViewCell & Reusable>(_ cellType: Cell.Type) { }
```
It's a generic because I don't want to enforce any specific type, it can be anything as long as it is a UITableViewCell and conforms to Reusable, so you can use any subclass you want here.

PR #5, opened 2022-07-11
Search

DiChizhik:


Services/MovieDataService.swift
```diff
@@ -133,6 +131,40 @@ class MovieDataService: MovieDataServiceProtocol {
...
+        var urlComponents = URLComponents()
+        urlComponents.scheme = "https"
+        urlComponents.host = "api.themoviedb.org"
+        urlComponents.path = "/3/search/movie"
+        urlComponents.queryItems = [
+            URLQueryItem(name: "api_key", value: "b2b14caf40262a9c19a366b15e4e3537"),
+            URLQueryItem(name: "query", value: query),
+            URLQueryItem(name: "page", value: "1")
+        ]
+        guard let url = urlComponents.url else { return }
```
GMinucci, line 144: Nicely done! I wonder how great this would look with the other methods we've implemented already 😛

ViewControllers/SearchViewController.swift
```diff
@@ -0,0 +1,123 @@
+//
+//  SearchViewController.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 11/07/2022.
+//
+
+import UIKit
+
+class SearchViewController: UIViewController {
```
GMinucci: Don't forget our architecture :)
View setup should be in a separate class
Also, about file structure, I'd recommend to put each "screen" and its related files under a single folder to help you find code that are related, for example:
```
\Movies (root of project)
    \... (other folders)
    \Screen (could also be "Feature" or "Journey" for example)
        \Playing Now
             PlayingNowCollectionViewController.swift
             PlayingNowCollectionViewCell.swift
        \Most Popular
             MostPopularViewController.swift
             MostPopularView.swift
             MostPopularCollectionViewCell.swift
        \Search
             ...    
        \Movie Details
             ...
```

ViewControllers/SearchViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+    private lazy var searchController: UISearchController = {
+       let search = UISearchController(searchResultsController: nil)
+        search.searchResultsUpdater = self
+        search.obscuresBackgroundDuringPresentation = false
+        search.automaticallyShowsCancelButton = true
+        search.hidesNavigationBarDuringPresentation = false
+        search.searchBar.placeholder = "Type movie name"
+        search.searchBar.tintColor = UIColor(named: "titleColor")
+//        This option doesn't work from here but works OK from loadView()
+//        search.searchBar.searchTextField.textColor = UIColor(named: "titleColor")!
```
GMinucci: I'm not entirely sure why, but when you set the `searchController` to the `navigationItem.searchController` the navigation item is applying its own style to some of the `searchController` styles which is why `searchTextField.textColor` changes to white. So, having it here works, it's just getting replaced later on.

ViewControllers/SearchViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        
+//        This option works but looks bulky and changes the size of the textField
+//        UITextField.appearance(whenContainedInInstancesOf: [UISearchBar.self]).defaultTextAttributes = [NSAttributedString.Key.foregroundColor: UIColor(named: "titleColor")]
```
GMinucci: Exactly, I'd advise against it and go with the approach you did, setting the color at a later time. I'd recommend using the Appearance protocol only in cases that you want to globally configure.

ViewControllers/SearchViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+    }()
+    
+    private lazy var tableView: UITableView = {
+       let tableView = UITableView()
+        tableView.translatesAutoresizingMaskIntoConstraints = false
+        tableView.backgroundColor = UIColor(named: "backgroundColor")
+        tableView.register(SearchTableViewCell.self)
+        tableView.rowHeight = 124
+        tableView.dataSource = self
+        tableView.delegate = self
```
GMinucci: Suggestion: Include this property to make the keyboard dismiss when you drag the table
```swift
tableView.keyboardDismissMode = .onDrag
```

ViewControllers/SearchViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+            tableView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
+            tableView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor)
+        ])
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        let notificationCenter = NotificationCenter.default
+        notificationCenter.addObserver(self, selector: #selector(makeRequest), name: UIResponder.keyboardWillHideNotification, object: nil)
```
GMinucci: I have 2 big comments on these lines:

1. More a curiosity:
Using the method `addObserver(_, selector:, name:, object:)` is usually more convenient as you don't need to worry about remove the observer when dealocating the main class. The NotificationCenter does that automatically if the observer got dealocated. That won't happen if you use the closure observer approach:
```swift
notificationCenter.addObserver(forName: UIResponder.keyboardWillHideNotification, object: nil, queue: .main) { [weak self] notification in
            guard let self = self else { return }
            self.makeRequest()
        }
```
If you do like this, even with the `[weak self]` you still end up with a retain here as this instance will be kept in the NotificationCenter observers list, so to fix this you should always remove the observer when needed (when view will disappear for example). Also, don't forget to hold the reference to remove later!
```swift
let notification = notificationCenter.addObserver(forName: UIResponder.keyboardWillHideNotification, object: nil, queue: .main) { ... }
// then later on
notificationCenter.removeObserver(notification)
```

2. Even though this approach accomplish the result we want, we are using the wrong mean. If the keyboard is displayed by any other mean: another text field, and alert that requires input, answering a push notification from the push directly.
Ideally we should be using the target actions available from the text field:
```swift
// .editingChanged is triggered on every character change. So each letter added or removed would trigger 
// a new search event. Commonly, if you want to use this target, we combine it with other operations such
// "debounce" which discard inputs while user is tapping and only trigger when user stops changing for more
// than X seconds. This avoids spamming the search with intermediary (useless) search results.
searchController.searchBar.searchTextField.addTarget(self, action: #selector(tals), for: .editingChanged)

// This one is straightforward, triggers when editing is done. This can be either when user taps on "search" button or pull the keyboard to dismiss. This is not triggered when tapping in "cancel" button!
searchController.searchBar.searchTextField.addTarget(self, action: #selector(makeRequest), for: .editingDidEnd)
```

ViewControllers/SearchViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+        let selectedMovieID = searchResults[indexPath.row].id
+        
+        let detailViewController = MovieDetailViewController(selectedMovieID: selectedMovieID)
+        let detailNavigationController = UINavigationController(rootViewController: detailViewController)
+        present(detailNavigationController, animated: true)
+    }
+}
+
+extension SearchViewController: UISearchResultsUpdating {
+    func updateSearchResults(for searchController: UISearchController) {
```
GMinucci: Since we manually implemented the search handling methods I'd recommend leaving this method empty and moving your search code (code that is currently inside `updateSearchResults` to its own method, perhaps to the `makeRequest` method itself.

The method `updateSearchResults` is automatically called on any change of the search bar text content so you don't need to add observers or add targets to the search bar. The downside is that you will receive any change so you end up needing to have a debounce code here.

ViewControllers/SearchViewController.swift
```diff
@@ -0,0 +1,123 @@
...
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        let notificationCenter = NotificationCenter.default
+        notificationCenter.addObserver(self, selector: #selector(makeRequest), name: UIResponder.keyboardWillHideNotification, object: nil)
+    }
+    
+    @objc private func makeRequest() {
```
GMinucci: Code encapsulation is not only about private/public, sometimes we can leak functionality through method names. In this case here, our method is called `makeRequest` which is related to how the service behaves or should behave but has nothing to do with the method action itself. Also, the name makes complete sense as we are requesting it through an API but what if we change that to a local database? In this case it won't make sense anymore.
Since this method is invoked when the user finishes editing, we could name it: `didFinishEditing` or `startSearching` for example.

GMinucci (changes requested):
Nicely done! 🔥
I left a couple comments in the search approach but it's nice to see you went straight to the harder implementation with the search in the navigation bar. Well done!

ViewControllers/SearchViewController.swift
DiChizhik: I thought about it and even created a separate file for SearchView but then deleted it since there was just tableView there.
Will fix that!

ViewControllers/SearchViewController.swift
DiChizhik: The reason why I added observers was that updateSearchResults didn't trigger multiple searches. Once I entered one letter, search was launched and never repeated.

Features/Search/SearchView.swift
```diff
@@ -0,0 +1,68 @@
...
+//  Movies
+//
+//  Created by Diana Chizhik on 12/07/2022.
+//
+
+import UIKit
+
+protocol SearchViewDelegate: AnyObject {
+    func startSearching(for text: String)
+}
```
GMinucci, line 12: Nice! 👏

Features/Search/SearchViewController.swift
```diff
@@ -0,0 +1,77 @@
...
+//
+//  Created by Diana Chizhik on 11/07/2022.
+//
+
+import UIKit
+
+class SearchViewController: UIViewController, SearchViewDelegate {
+    let movieDataService = MovieDataService()
+    private var searchResults = [Movie]()
+    private var isSearching = false
```
GMinucci, line 13: I think we don't need this anymore

GMinucci (approved):
Just one very minor comment. Overall super good 😁

Sidenote: Don't forget about this comment [here](https://github.com/DiChizhik/Movies/pulls#discussion_r918872015). We can do it in a diff PR as well, no worries. It's always good to refactor your code when you see something that can be improved. Doing it gradually is easier than a big refactor of the whole app.

ViewControllers/SearchViewController.swift
GMinucci: That's actually a good discussion point. When you follow an architecture it's common that you find yourself on situations that you need to add lots of "unnecessary" files just to fulfill the architecture, we refer to that as boilerplate files.
It can be annoying to do it, but at same time, what's the purpose of the architecture if we follow it only when it's convenient 🤓
Since we are talking on architecture, I'll try to show you some of the common ones on iOS in our session :)

PR #6, opened 2022-07-18
Error handling

DiChizhik:


Features/ErrorViews/ErrorView.swift
```diff
@@ -0,0 +1,106 @@
...
+//  ErrorView.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 18.07.22.
+//
+
+import UIKit
+
+protocol ErrorViewDelegate: AnyObject {
+    func dismissController()
```
GMinucci: I wasn't being picky on this but the time has come 😄 It's recommended to pass the object as the first parameter of delegates, usually ommiting the name. In this case this would be:
```suggestion
    func dismissController(_ errorView: ErrorView)
```
Later, it would be called as:
```swift
delegate?.dismissController(self)
```

Features/DetailView/MovieDetailViewController.swift
```diff
@@ -61,7 +61,9 @@ class MovieDetailViewController: UIViewController {
                 }
             case .failure(let error):
                 DispatchQueue.main.async {
-                    self.showError(message: error.localizedDescription)
+                    let vc = ErrorViewController(error: error)
+                    vc.modalPresentationStyle = .popover
```
GMinucci: How this is looking like if you get an error while in the movie details screen?
Bonus: What happens if you present the error while running the app in an iPad? I'm not saying this for you to always test everything in the iPad as well, I'm just asking because we should be using a different presentation style here

Features/ErrorViews/ErrorView.swift
```diff
@@ -0,0 +1,106 @@
...
+        let view = UIStackView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.axis = .vertical
+        view.spacing = 24
+        view.alignment = .center
+        view.layoutMargins = UIEdgeInsets(top: 24, left: 32, bottom: 16, right: 32)
+        view.isLayoutMarginsRelativeArrangement = true
+        view.backgroundColor = UIColor(named: "errorViewBackgroundColor")
+        view.layer.cornerRadius = 5
+        view.layer.masksToBounds = true
```
GMinucci, line 60: Curiosity: In the past (before iOS 14) this wouldn't work. UIStackViews are purely structural meaning you can only use them for arranging views. Under the hood it used to be a `CATransformLayer`. After iOS 14 it got changed to a `CALayer` which can now handle UI properties. More [here](https://useyourloaf.com/blog/stack-view-background-color-in-ios-14/).

Features/ErrorViews/ErrorView.swift
```diff
@@ -0,0 +1,106 @@
...
+            stackView.widthAnchor.constraint(equalTo: safeAreaLayoutGuide.widthAnchor, multiplier: 0.76),
+            stackView.heightAnchor.constraint(lessThanOrEqualTo: safeAreaLayoutGuide.heightAnchor, multiplier: 0.8)
+        ])
+        
+        stackView.addArrangedSubview(imageView)
+        stackView.addArrangedSubview(errorLabel)
+        stackView.addArrangedSubview(buttonStackView)
+        NSLayoutConstraint.activate([
+            buttonStackView.trailingAnchor.constraint(equalTo: stackView.layoutMarginsGuide.trailingAnchor)
+        ])
```
GMinucci: When using stack views try avoiding setting constraints to views that are inside of the stack, this can go to some weird directions sometimes. Stacks can be super annoying to deal so what to do here: 

1. First, usually you will always set your stack alignment to `.fill`, specially on high level stacks. So in our case here we need to set the high level stack alignment to `.fill`.

2. Secondly, our image view will try to fill it ignoring aspect ratio, since we need it centered we make `imageView.contentMode = .center`

3. Lastly, your approach for positioning the button is super on point! Already prepared to handle multiple buttons as well so congrats! We just need to achieve this behaviour (pushing to the right) without a constraint. Since our main stack is now `.fill` mode, the button stack will fill the full space, to move the button to the right most we simply add a spacer to the beginning. In this case a simple UIView can do the trick since buttons by default hug themselves to the minimum space already, the UIView will simply take as much space available as possible pushing the button all the way to the right. Just don't forget to update the buttons stack axis to `.horizontal`, alignment to `.fill`, and remove the manual constraints

Features/ErrorViews/ErrorViewController.swift
```diff
@@ -0,0 +1,55 @@
...
+            contentView.errorLabel.text = "Houston, we have a problem.\nClose and re-open the app."
+        case .failedToDecode:
+            contentView.imageView.image = UIImage(named: "spy")
+            contentView.errorLabel.text = "Failed to decode data.\nCall the special agents."
+        }
+    }
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+    }
```
GMinucci: If we don't use it, we burn it 🔥

Features/ErrorViews/ErrorViewController.swift
```diff
@@ -0,0 +1,55 @@
...
+        case .failedToGetResponse:
+            contentView.imageView.image = UIImage(named: "wifi")
+            contentView.errorLabel.text = "Sherlock didn’t find the internet signal.\nPlease try again later."
+        case .failedToGetData:
+            contentView.imageView.image = UIImage(named: "dizzy")
+            contentView.errorLabel.text = "Houston, we have a problem.\nClose and re-open the app."
+        case .failedToDecode:
+            contentView.imageView.image = UIImage(named: "spy")
+            contentView.errorLabel.text = "Failed to decode data.\nCall the special agents."
+        }
```
GMinucci: With this approach, on every new error added we would have to manually come here and add the proper handling. Do you see any different approach? Perhaps a way to handle any current or future errors.
Bonus: Can we handle any kind of error, even those who are not `MovieServiceError`?

Features/MostPopular/MostPopularViewController.swift
```diff
@@ -41,7 +41,10 @@ class MostPopularViewController: UIViewController, MostPopularViewDelegate {
         
         loadMovieData()
         
+//        The app crashed if there were no movies(Error: Failed to load from server), so I added this check to prevent the crash.
         DispatchQueue.main.async {
+            guard !self.movies.isEmpty else { return }
```
GMinucci, line 46: This will solve the issue, but I want to understand exactly what is causing the crash. Can you dig a bit to narrow where the crash is happening? If this is not easily reproducible you can add a small delay to the API response: go to `MovieDataService`, at line 75, replace it with this:
```swift
DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
    completion(.success(moviesData.results))
}
```
It will delay the success response by 1 second.

Features/PlayingNow/PlayingNowCollectionViewController.swift
```diff
@@ -43,8 +43,12 @@ class PlayingNowCollectionViewController: UICollectionViewController {
...
             case .success(let moviesList):
                 self.movies.append(contentsOf: moviesList)
-            case .failure(_):
-                break
+            case .failure(let error):
+                DispatchQueue.main.async {
+                    let vc = ErrorViewController(error: error)
+                    vc.modalPresentationStyle = .popover
+                    self.present(vc, animated: true)
+                }
```
GMinucci: What about wrapping this in a method to be easily called from any place? Maybe:
```swift
class ErrorViewController {
    static func handleError(_ error: MovieServiceError, presentingViewController: UIViewController) { ... }
}
```
Then in any place you want to use we can simply call:
```swift
ErrorViewController.handleError(error, presentingViewController: self)
```

Models/MovieDetails.swift
```diff
@@ -32,7 +32,12 @@ struct MovieDetails: Codable {
         title = try values.decode(String.self, forKey: .title)
         
         if let posterPathString = try values.decodeIfPresent(String.self, forKey: .posterPath) {
-            posterPath = URL(string: "https://image.tmdb.org/t/p/w500" + posterPathString)
+            var urlComponents = URLComponents()
+            urlComponents.scheme = "https"
+            urlComponents.host = "image.tmdb.org"
+            urlComponents.path = "t/p/w500\(posterPathString)"
+            
+            posterPath = urlComponents.url
```
GMinucci: Super 🔥

Movies.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -5,8 +5,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/onevcat/Kingfisher.git",
       "state" : {
-        "revision" : "022e4eeb79f817748544b318b991d9a70036bbf8",
-        "version" : "7.2.4"
```
GMinucci, line 9: Just a FYI here, the package version is set to `7.0.0 - Next Major` which means it will try to update to latest version until next major (8.0.0). Usually only major versions will contain a breaking change when upgrading, and the key word here is "usually" 😁
If you want to be on the safest side possible it's good to lock the version to exact one number, that way you are sure to only upgrade when you are really meaning to upgrade. Downside is that you loose smaller upgrades and improvements because you now need to do that manually.

Services/MovieDataService.swift
```diff
@@ -38,7 +38,16 @@ class MovieDataService: MovieDataServiceProtocol {
...
     func getPlayingNowMoviesList(completion: @escaping (Result<[Movie], MovieServiceError>)-> Void) {
         guard isPlayingNowRequestCompleted else { return }
-        guard let url = URL(string: "https://api.themoviedb.org/3/movie/popular?api_key=b2b14caf40262a9c19a366b15e4e3537&language=en-US&page=\(playingNowPage)") else { return }
+        
+        var urlComponents = URLComponents()
+        urlComponents.scheme = "https"
+        urlComponents.host = "api.themoviedb.org"
+        urlComponents.path = "/3/movie/now_playing"
+        urlComponents.queryItems = [
+            URLQueryItem(name: "api_key", value: "b2b14caf40262a9c19a366b15e4e3537"),
```
GMinucci: Can we move the API Key to be in a single place? Maybe a private static property, that way we are not risking a typo or in case we need to change the key it will be in a single place.

GMinucci (changes requested):
Great job! Nice to see that you are following our architecture even on "not clearly a new screen" places. I've added a couple comments to be addressed and some are just curiosities or FYI's

Features/DetailView/MovieDetailViewController.swift
DiChizhik: I've changed the presentation style to .pageSheet. It works in both cases.

.popover worked exactly the same with DetailVC, but crashed on iPad.

Features/ErrorViews/ErrorView.swift
DiChizhik: I feel lucky to have all this now :)
So, back then I'd have to use UIStackView with another  UIView underneath it?

Features/ErrorViews/ErrorView.swift
DiChizhik: Beautiful, thanks!
Stacks are tricky beasts :)

Features/ErrorViews/ErrorViewController.swift
DiChizhik: I'll extend MovieServiceError enum with image property.

We can since MovieServiceError conforms to Error protocol. So, I'll use Error protocol as type and pass it as parameter type to ErrorVC init. Then downcast to handle known errors.

Features/MostPopular/MostPopularViewController.swift
DiChizhik: It crashed because the vc has to scroll to a certain item in the array. 
`var itemInViewIndex: Int = 0`
`self.contentView.collectionView.scrollToItem(at: IndexPath(item: self.itemInViewIndex, section: 0), at: UICollectionView.ScrollPosition.centeredHorizontally, animated: false)`

 Since there are no items, it crashed.

Features/PlayingNow/PlayingNowCollectionViewController.swift
DiChizhik: Beautiful!

I was thinking about it but didn't work out the way to implement.

Services/MovieDataService.swift
DiChizhik: Should I do the same with host?

I've also seen people add a separate enum or struct with constants. Is it a good idea?

Features/ErrorViews/ErrorView.swift
GMinucci: Yep, exactly

Services/MovieDataService.swift
GMinucci: Yep, usually those are defined in a "configuration" place which can be a config file (json, plist) or a swift file with constants. You can create structs or enums to separate them as well, to define a "namespace" to organize it. When creating namespaces I prefer going with an enum instead of a struct, but both would work.

GMinucci (approved):
Super 😄

PR #7, opened 2022-07-25
Watchlist

DiChizhik:
I've implemented the logic for watchlist adding/deleting only in MostPopularViewController.

There's also a lot of inconsistencies in the implementation of buttons and views since I've tried several ways of making them and would appreciate your feedback on them.

What I mean, I know this version is very raw, I just would like to ensure I'm not doing something totally wrong.


Features/DetailView/DetailCollectionViewCell.swift
```diff
@@ -49,3 +40,17 @@ class DetailCollectionViewCell: UICollectionViewCell, Reusable {
...
+private extension DetailCollectionViewCell {
+    func setupUI() {
+        contentView.addSubview(label)
+        
+        NSLayoutConstraint.activate([
+            label.topAnchor.constraint(equalTo: contentView.topAnchor, constant: 4),
+            label.leadingAnchor.constraint(equalTo: contentView.leadingAnchor, constant: 15),
+            label.trailingAnchor.constraint(equalTo: contentView.trailingAnchor, constant: -15),
+            label.bottomAnchor.constraint(equalTo: contentView.bottomAnchor, constant: -4),
+        ])
```
GMinucci: We have lot of places where this same piece of code happens. Setting constraints to another view (usually the superview) around all edges. It would be nice if we develop our own "constraints helper" to make this less repetitive. 🤔

Features/DetailView/MovieDetailViewController.swift
```diff
@@ -89,7 +99,7 @@ class MovieDetailViewController: UIViewController {
         contentView.collectionView.reloadData()
     }
     
-    private func showError(message: String) {
+    func showError(message: String) {
```
GMinucci: Let's refactor this to use our new error handling method

Features/DetailView/StackMovieDetailView.swift
```diff
@@ -143,23 +161,27 @@ class StackMovieDetailView: UIView {
...
         centerAlignedStack.addArrangedSubview(titleLabel)
         centerAlignedStack.addArrangedSubview(imageView, spaceBefore: 16)
+        centerAlignedStack.addArrangedSubview(reviewAndWatchlistStack, spaceBefore: 19)
         
         let releaseStack = UIStackView(arrangedSubviews: [releasedOnLabel, releaseDateLabel])
-        releaseStack.translatesAutoresizingMaskIntoConstraints = false
         releaseStack.axis = .horizontal
-        releaseStack.alignment = .leading
+//        Why do we need .leading here if its description says "a layout for vertical stacks where the stack view aligns the leading edge of its arranged views along its leading edge. This is equivalent to the top alignment for horizontal stacks."
+//        releaseStack.alignment = .leading
```
GMinucci, line 187: Not sure if I got the question. What was the issue? Did it push one of the labels all the way to the right?

Features/DetailView/StackMovieDetailView.swift
```diff
@@ -170,7 +192,9 @@ class StackMovieDetailView: UIView {
         
         NSLayoutConstraint.activate([
             imageView.widthAnchor.constraint(equalTo: scrollContentStack.widthAnchor, multiplier: 0.61),
-            imageView.heightAnchor.constraint(equalTo: imageView.widthAnchor, multiplier: 1.41)
+            imageView.heightAnchor.constraint(equalTo: imageView.widthAnchor, multiplier: 1.41),
+            
+            reviewAndWatchlistStack.widthAnchor.constraint(equalTo: imageView.widthAnchor)
```
GMinucci, line 205: I don't think there is an issue here to set the stack width constraint, if you want the stack to always have the same width as the image that's completely fine. You can get issues when you set positioning constraints to arranged subviews of a stack

Features/LaunchViewController.swift
```diff
@@ -7,6 +7,7 @@
 
 import UIKit
 
+//Shall I create a separate view for this controller to comply with MVC?
```
GMinucci: Ideally... yes :)
I know it's super annoying, specially with really small screens. But imagine we want to change the splash to start pre-loading data, reading user ui preferences, doing some crazy animations and so on. If the base is ready it's easier to maintain and improve.

Features/LaunchViewController.swift
```diff
@@ -59,7 +63,7 @@ class LaunchViewController: UIViewController {
                         })
     }
     
-    private func presentViewController() {
+    func presentViewController() {
         let tabBarController = UITabBarController()
```
GMinucci: Some future improvements. What if instead of using the vanilla UITabBarController we create our own MovieAppTabBarController (or any other name) which will have all of this configuration done by it: Tab bar items, tab bar styles, view controllers, etc. This way we can take away this responsibility from the splash screen.

Features/MostPopular/MostPopularView.swift
```diff
@@ -9,10 +9,30 @@ import UIKit
...
+    }
+    
+    weak var collectionViewDataSource: UICollectionViewDataSource? {
+        get {
+            collectionView.dataSource
+        }
+        set {
+            collectionView.dataSource = newValue
+        }
+    }
```
GMinucci: So the idea here is that you can keep the collection view private while exposing its delegate and datasource. Interesting stuff!

Features/MostPopular/MostPopularView.swift
```diff
@@ -24,62 +44,62 @@ class MostPopularView: UIView {
...
     }()
     
-    private(set) lazy var reviewsScoreIndicator: UIImageView = {
-        let reviewsScoreIndicator = UIImageView()
-        reviewsScoreIndicator.translatesAutoresizingMaskIntoConstraints = false
-        reviewsScoreIndicator.heightAnchor.constraint(equalToConstant: 15).isActive = true
-        reviewsScoreIndicator.widthAnchor.constraint(equalTo: reviewsScoreIndicator.heightAnchor, multiplier: 1.13).isActive = true
-        reviewsScoreIndicator.alpha = 0
-        return reviewsScoreIndicator
+//    As of now it's the only place where I've tried out UIButton extension method and used WatchlistService.
```
GMinucci: I was trying to understand why it wasn't working for me 😂 now I get it haha

Features/MostPopular/MostPopularView.swift
```diff
@@ -24,62 +44,62 @@ class MostPopularView: UIView {
...
+    private(set) lazy var watchlistButton: UIButton = {
+        let button = UIButton.createWatchlistButton()
+        button.translatesAutoresizingMaskIntoConstraints = false
+        button.alpha = 0
+        let action = UIAction { [weak self] _ in
+            self?.delegate?.watchlistTapped()
+        }
```
GMinucci: Ok, this is new to me. It seems it was introduced in iOS 13. Super cool! 🔥

Features/MostPopular/MostPopularView.swift
```diff
@@ -24,62 +44,62 @@ class MostPopularView: UIView {
...
         description.font = UIFont.systemFont(ofSize: 15)
-        description.textColor = UIColor(named: "descriptionColor")
+        description.textColor = .pureWhiteFF
         description.textAlignment = .center
         description.alpha = 0
         return description
     }()
     
-    private(set) lazy var seeMoreButton: UIButton = {
+//    If I use buttonConfiguration in one place, is it critical I use it with all buttons?
```
GMinucci: Not sure if I understood what is buttonConfiguration. Is it the extension you created with `UIButton.createWatchlistButton`?
If that's the case, I'd say that's not critical. You usually have that type of builder for components that are used very often. Things that only exists on one screen it makes sense to only create on that context. As a rule of thumb, if you use the same component more than 3 times, consider making it reusable with a builder or an easy configure method.

Features/MostPopular/MostPopularView.swift
```diff
@@ -89,9 +109,43 @@ class MostPopularView: UIView {
...
         setupUI()
     }
- 
-    private func setupUI() {
-        backgroundColor = UIColor(named: "backgroundColor")
+}
+
+// MARK: - Public functions
+extension MostPopularView {
+    func configureWith(name: String, score: Int, description: String?, status: WatchlistStatus?) {
```
GMinucci, line 122: Just a quick tip: with configure methods, as the view gets bigger it can happen that your configure method starts to need lots of parameters. When that happens you can wrap the parameters into a convenient struct. As a rule of thumb, more than 4 parameters we can start grouping them. Taking this one as an example (not that you need to do it for this one!)
```swift
struct MostPopularViewConfiguration {
    let name: String
    let score: Int
    let description: String?
    let status: WatchlistStatus?
}

func configure(with configuration: MostPopularViewConfiguration) { ... }
```

Features/MostPopular/MostPopularView.swift
```diff
@@ -89,9 +109,43 @@ class MostPopularView: UIView {
...
+        if let status = status {
+            switch status {
+            case .added:
+                watchlistButton.updateWatchlistButton(isOnWatchlist: true)
+            case .notAdded:
+                watchlistButton.updateWatchlistButton(isOnWatchlist: false)
+            }
+        } else {
+            watchlistButton.updateWatchlistButton(isOnWatchlist: false)
+        }
```
GMinucci: `status` being optional implies we have always one extra state besides the ones in the enum. Does it make sense for our architecture? Is it better if we ensure that this parameter is always passed and never optional? If we choose to keep it optional, we can use the `case .none` to make this a bit more readable:
```swift
switch status {
    case .added:
        // when added
            
    case .notAdded:
        // when not added
            
    case .none:
        // when nil
}
```

Features/MostPopular/MostPopularViewController.swift
```diff
@@ -65,55 +97,32 @@ class MostPopularViewController: UIViewController, MostPopularViewDelegate {
         }
     }
     
-    private func configureWithData(index: Int) {
+    func configureWithData(index: Int) {
         let movie = movies[index]
```
GMinucci: Don't forget about our brand new `movies[safe: index]`

Features/PlayingNow/PlayingNowCollectionViewCell.swift
```diff
@@ -22,78 +22,78 @@ class PlayingNowCollectionViewCell: UICollectionViewCell, Reusable {
...
-    private lazy var reviewsScoreIndicator: UIImageView = {
-        let reviewsScoreIndicator = UIImageView()
-        reviewsScoreIndicator.translatesAutoresizingMaskIntoConstraints = false
-        reviewsScoreIndicator.heightAnchor.constraint(equalToConstant: 15).isActive = true
-        reviewsScoreIndicator.widthAnchor.constraint(equalTo: reviewsScoreIndicator.heightAnchor, multiplier: 1.13).isActive = true
-        return reviewsScoreIndicator
+    private lazy var watchlistButton: WatchlistButton = {
+        let button = WatchlistButton()
+        button.translatesAutoresizingMaskIntoConstraints = false
+//      Need to figure out how to get rid of the right inset. Might disappear when implemented using UIButton extension method. Maybe create a separate one for the short version of the button?
```
GMinucci: UIButtons usually have some weird insets that are a bit annoying to deal depending on what you are trying to achieve. There are a couple of ways to do it easier:
 - Subclass UIView directly, so you can use the `touchesBegan`, `touchesEnded` and `touchesCanceled` to handle user interaction. A bit harder but gives you complete control on the view behaviour and touch experience
 - Subclass UIControl. This one is easier that subclassing UIView, allows you do usual things you do with a button like `addTarget` and a bit more convenient to deal when customizing UI. UIButton is a subclass of UIControl. Note that when comes to accessibility you need to provide all traits and descriptions in order for this to work. The button will do most of it for free.

Features/WatchList/WatchlistTableViewCell.swift
```diff
@@ -0,0 +1,12 @@
+//
+//  WatchlistTableViewCell.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 21.07.22.
+//
+
+import UIKit
+
+class WatchlisttableViewCell: SearchTableViewCell {
```
GMinucci: Just a small typo
```suggestion
class WatchlistTableViewCell: SearchTableViewCell {
```
I like the idea of subclassing so you can reuse the component and keep different names. But which one is better? We create another class as a subclass or should we make the first a reusable component. Instead of `SearchTableViewCell` maybe just `MovieListTableViewCell`. This is not a request to change, I jsut want you to think about which one would be better in your opinion, and why

Features/WatchList/WatchlistViewController.swift
```diff
@@ -0,0 +1,77 @@
...
+                    self?.movies.append(details)
+                    DispatchQueue.main.async { [weak self] in
+                        self?.contentView.tableView.reloadData()
+                    }
+                case .failure(_):
+                    break
+                }
+            }
+        }
+    }
```
GMinucci: We are fetching movie details from the API for every movie in the list. Imagine we have 100 movies in our list, we do 100 requests when entering this screen. Can we optimize this? Do you see any architecture change that would make this better?

Models/WatchlistItem.swift
```diff
@@ -0,0 +1,13 @@
...
+//
+//  Created by Diana Chizhik on 21.07.22.
+//
+
+import Foundation
+
+struct WatchlistItem: Codable {
+    var id: Int
+    var saveDate: Date
+}
```
GMinucci: Nice to see we are opting to store the bare minimum! That's a good practice so we keep our database size as small as possible while showing to the user everything he needs.

Services/WatchlistService.swift
```diff
@@ -0,0 +1,86 @@
...
+class WatchlistService {
+    typealias Watchlist = [Int : WatchlistItem]
+    
+    func getStatus(for id: Int)-> WatchlistStatus? {
+        if let watchlist = getWatchlist() {
+            return watchlist[id] != nil ? .added : .notAdded
+        } else {
+            return nil
+        }
+    }
```
GMinucci: What if we make `getWatchlist` return an empty dictionary in case nothing is stored instead of `nil`? Then this method can be a lot more simpler and we get rid of the `nil` status which seems weird. If `.notAdded` means that is not added yet, what does `nil` means?

```suggestion
    func getStatus(for id: Int)-> WatchlistStatus {
        let watchlist = getWatchlist()
        return watchlist[id] != nil ? .added : .notAdded
    }
```

Services/WatchlistService.swift
```diff
@@ -0,0 +1,86 @@
...
+    
+    func toggleStatus(for item: WatchlistItem)-> WatchlistStatus? {
```
GMinucci: Same as the previous comment, a `nil` `WatchListStatus` sounds weird. With the change I suggested to `getWatchlist` we can get rid of that and improve the readability of the method here as well as we don't need to handle 3 states anymore.

Services/WatchlistService.swift
```diff
@@ -0,0 +1,86 @@
...
+                return saveWatchlist(watchlist) ? .added : nil
+            }
+        } else {
+            let watchlist: Watchlist = [item.id : item]
+            return saveWatchlist(watchlist) ? .added : nil
+        }
+    }
+    
+    func getWatchlistItems() -> [WatchlistItem]? {
+        guard let watchlist = getWatchlist() else { return nil }
```
GMinucci: Why do we return `nil` when there is no movie in the watchlist? Shouldn't it be just an empty list instead?

Services/WatchlistService.swift
```diff
@@ -0,0 +1,86 @@
...
+            items.append(value)
+        }
+
+        return items
+    }
+    
+    private func getWatchlist() -> Watchlist? {
+        let defaults = UserDefaults.standard
+        
+        if let watchlistData = defaults.object(forKey: "watchlist") as? Data {
```
GMinucci: We need to be careful with strings that are keys to identify objects. This can be a reuse identifier for a table view cell or a name of a property in a user preferences. We can easily put a typo and mess everything. Try to always put it in a single static property and always refer to that property instead of rewriting the string.

UIButton.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  UIButton.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 20.07.22.
+//
+
+import UIKit
+
+extension UIButton {
```
GMinucci: Nice! It's good to extend base classes to include some convenient methods.

UIButton.swift
```diff
@@ -0,0 +1,62 @@
...
+        )
+        self.titleEdgeInsets = UIEdgeInsets(
+            top: 0,
+            left: imageTitlePadding,
+            bottom: 0,
+            right: -imageTitlePadding
+           )
+    }
+    
+    static func createWatchlistButton()-> UIButton {
```
GMinucci: I like this pattern and I've seem a lot of developers using it!
Only suggestions is that I personally like to keep things super organized so I'd put this in a namespace to tidy a bit more the Xcode autocompletion. But, this is just a personal taste 😉 
Maybe:

```swift
extension UIButton { 
    enum MovieApp { 
        static func createWatchListButton() -> UIButton { ... }
    }
}

// And to use it
let button = UIButton.MovieApp.createWatchlistButton()
```

UIButton.swift
```diff
@@ -0,0 +1,62 @@
...
+            var outgoing = incoming
+            outgoing = .lightBlue61
+            return outgoing
+        }
+
+        let button = UIButton(configuration: config)
+        return button
+    }
+    
+    func updateWatchlistButton(isOnWatchlist: Bool) {
```
GMinucci: This one I would be against it. This is customizing the button for an specific scenario in a single case, and the code exists in an `UIButton` extension which means that all buttons will have access to this method. This should be done in a place that only concerns the class of buttons that this code makes sense to.

UIColor.swift
```diff
@@ -0,0 +1,18 @@
...
+
+extension UIColor {
+    static let lightBlue61 = UIColor.init(named: "lightBlue61") ?? .blue
+    static let green = UIColor.init(named: "green") ?? .green
+    static let whiteF5 = UIColor.init(named: "whiteF5") ?? .white
+    static let darkBlue01 = UIColor.init(named: "darkBlue01") ?? .black
+    static let pureWhiteFF = UIColor.init(named: "pureWhiteFF") ?? .white
+    static let blue2A = UIColor.init(named: "blue2A") ?? .blue
+    static let lightBlueDB = UIColor.init(named: "lightBlueDB") ?? .lightGray
+}
```
GMinucci: Oh la la! Great stuff 😎 
Note: you can omit the init part:
```swift
static let lightBlue61 = UIColor(named: "lightBlue61") ?? .blue
```

Views/WatchlistButton.swift
```diff
@@ -0,0 +1,67 @@
+//
+//  WatchListButton.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 20.07.22.
+//
+
+import UIKit
+// I planned to delete this implementation of the button and replace it with methods from the UIButton extension. However, it'd be great to discuss both.
```
GMinucci: Which one do you prefer? Both are valid ways of doing it.

GMinucci (changes requested):
Great stuff here! I've left some comments to be addressed and I have a couple things we need to do before moving on to CoreData. Let's make this fully ready so when we work with CoreData is just a matter of swapping the services.
To be done:
- If you open watchlist, then go add a movie to watchlist, then go back to watchlist tab it will never refresh the list for you.
- Let's make it possible to add to watchlist from everyplace we have a button for that :)

Bonus points: If you put a movie in watch list, that movie should reflect it from everywhere. So you add it in the Most Popular screen? Switching back to playing now should show that is in the watchlist automatically.

Features/DetailView/StackMovieDetailView.swift
DiChizhik: I meant that .leading I defined as a layout for vertical stacks, not horizontal. it seems to have no impact when I comment it altogether.

Features/WatchList/WatchlistTableViewCell.swift
DiChizhik: I guess the cons of subclassing is that there will be issues if we decide to modify the parent class.
And `MovieListTableCell ` could be a template that both of these cells will inherit from. Then we'll be free to modify the subclasses.

Features/DetailView/MovieDetailViewController.swift
DiChizhik: I'm not sure how to do it now, since this branch watch created from the old main version, the one that existed before the merge with errorHandling

Features/DetailView/StackMovieDetailView.swift
DiChizhik: I see. 
So, here I don't really set constraints to the width of the image since the width of the image itself is set according to the width of the stack. Did I get that right?

Features/MostPopular/MostPopularView.swift
DiChizhik: Yeah, no more addTarget and #selector.
I'll try changing to UIAction for seeMoreButton as well.

Features/MostPopular/MostPopularView.swift
DiChizhik: I rather meant the different API for UIButton, the old and the new. I guess it's better to be consistent about using them, right?

Features/MostPopular/MostPopularView.swift
DiChizhik: Looks niiiiice!

So, if I use it, I put it inside MostPopularView class?

Features/MostPopular/MostPopularView.swift
DiChizhik: I don't think it makes sense really to add` .none`. The status is either `.added` or `.notAdded.` I'll fix it.

Features/DetailView/MovieDetailViewController.swift
DiChizhik: I've rebased my first branch!
And it seems to have fixed itself :)

Features/WatchList/WatchlistViewController.swift
DiChizhik: There are several things we could do:
1) As discussed in our last call, we can save all movie data except posters in CoreData. Then get them using a separate method to get images in `MovieDataService` https://developers.themoviedb.org/3/movies/get-movie-images
Could we actually save posters too?

2) We could try enabling pagination to split the requests into groups of 10 for example. 

3) I also recall that there's NSFetchedResultsController in CoreData with adjustable fetchBatchSize. However, it's not related to API requests any more.

Services/WatchlistService.swift
DiChizhik: Got it!

UIColor.swift
DiChizhik: Sure, my bad

Services/WatchlistService.swift
DiChizhik: We get nil here if there is no watchlist. Is it better to return an empty array?

Features/DetailView/MovieDetailViewController.swift
GMinucci: Yep, good one!
FYI: there are two ways to solve this:
1. Rebase your current branch, so you get all latest changes from develop
2. Merge develop on top of your current branch
Both ways will have the same outcome (you get the latest code). I personally prefer the rebase method, I don't like how merging produces an extra commit. And, technically speaking, your operation is rebasing so why using a merge to perform a rebase? 😄

Features/DetailView/StackMovieDetailView.swift
GMinucci: Some properties can behave a bit different depending on the axis (horizontal/vertical). I don't know if you ever found this page but this helped me a lot (and still helps sometimes) to understand stacks properly: https://nshipster.com/uistackview/

Features/DetailView/StackMovieDetailView.swift
GMinucci: When the alignment is set to `.fill` the cross axis (the opposite of the axis) will take the full size. So:
Horizontal stack with fill alignment: subviews will have the stack height as its height. 
Vertical stack with fill alignment: subviews will have the stack width as its widht.

Features/MostPopular/MostPopularView.swift
GMinucci: Yep, that's right idd. Keep in mind we need to think always "is this a component or something I'll add only for this screen?", with that you can decide if you should make a configuration or simply customize in place.

Features/MostPopular/MostPopularView.swift
GMinucci: I'm a big fan of an organized namespace, so I'd like that 😄 
Additionally, if it's inside `MostPopularView` class you don't need this big name, simply `ViewConfiguration` would work, otherwise you need to declare it as `MostPopularView.MostPopularViewConfiguration` as opposed to `MostPopularView.ViewConfiguration`

Features/MostPopular/MostPopularView.swift
GMinucci: Just to be sure here, `.none` is not an state on status but rather a state on any property that is optional.
```swift
enum MyEnum {
    case one, two
}

let optionalEnum: MyEnum? = .one
switch optionalEnum {
    case .one:
        break
    case .two:
        break
    case .none:
        break
}

let nonOptionalEnum: MyEnum = .one
switch optionalEnum {
    case .one:
        break
    case .two:
        break
}
```

Features/WatchList/WatchlistTableViewCell.swift
GMinucci: Perfect!

Services/WatchlistService.swift
GMinucci: That's an architectural question, if our app only allow 1 watchlist and we make that everyone has a watchlist that is empty if no movie was added we don't have to deal with `nil` lists. But again, the choice is yours :)

Features/MostPopular/MostPopularView.swift
```diff
@@ -89,9 +115,41 @@ class MostPopularView: UIView {
...
+    
+    func changeAlpha(to value: CGFloat) {
+        name.alpha = value
+        reviewScoreStackView.alpha = value
+        watchlistButton.alpha = value
+        movieDescription.alpha = value
+        seeMoreButton.alpha = value
+    }
+    
+//    I've added this function so that we don't need to access the watchlistButton itself to call its updateWithStatus function. However, due to the Handleable protocol I can't make the button private. In this case does this function make sense here?
```
GMinucci, line 143: I see, I think the issue here is due to how the `WatchlistHandleable` is implemented. Let's discuss it in our sync.

Features/MovieTabBar/MovieTabBarController.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  MovieTabBarController.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 3.08.22.
+//
+
+import UIKit
+
+class MovieTabBarController: UITabBarController {
```
GMinucci, line 10: oh la la! ❤️

Features/MovieTabBar/MovieTabBarController.swift
```diff
@@ -0,0 +1,62 @@
...
+            let vc = item.rootViewController
+            let navController = UINavigationController(rootViewController: vc)
+            navController.tabBarItem.title = item.title
+            navController.tabBarItem.image = item.image
+            
+            controllers.append(navController)
+        }
+        
+        viewControllers = controllers
+        assert(viewControllers != nil)
```
GMinucci: Just curious why the assert?

Features/MovieTabBar/MovieTabBarItem.swift
```diff
@@ -0,0 +1,49 @@
+//
+//  MovieTabBarItem.swift
+//  Movies
+//
+//  Created by Diana Chizhik on 3.08.22.
+//
+
+import Foundation
+import UIKit
```
GMinucci, line 9: If you import UIKit you don't need to import Foundation as well
```suggestion
import UIKit
```

Features/MovieTabBar/MovieTabBarItem.swift
```diff
@@ -0,0 +1,49 @@
...
+
+enum MovieTabBarItem: String, CaseIterable {
```
GMinucci, line 11: This is super nice! 🥇

Features/MovieTabBar/MovieTabBarItem.swift
```diff
@@ -0,0 +1,49 @@
...
+        case .mostPopular:
+            return #imageLiteral(resourceName: "mostRecentIcon")
+        case .watchlist:
+            return #imageLiteral(resourceName: "ticket")
+        case .search:
+            return UIImage(systemName: "magnifyingglass.circle")!
+        }
+    }
+    
+    var rootViewController: UIViewController {
```
GMinucci, line 34: I like this one too but keep in mind that every time you call it, a new instance will be created. This won't have any impact on this app for now, but if we implement deep linking for example we may have issues if we don't keep this in mind.

Features/PlayingNow/PlayingNowCollectionViewController.swift
```diff
@@ -54,33 +63,49 @@ class PlayingNowCollectionViewController: UICollectionViewController {
...
-        super.viewWillTransition(to: size, with: coordinator)
-        collectionView.collectionViewLayout.invalidateLayout()
+}
+
+// MARK: - WatchlistButtonDelegate
+extension PlayingNowCollectionViewController: WatchlistButtonDelegate {
+    func watchlistTapped(_ view: WatchlistHandleable) {
+        guard let cell = view as? UICollectionViewCell else { return }
+        
+        if let indexPath = collectionView.indexPath(for: cell) {
```
GMinucci, line 89: Having an index path is crucial for this method to execute so it makes more sense to move this to a guard statement

Features/WatchList/WatchlistViewController.swift
```diff
@@ -0,0 +1,104 @@
...
+        
+        title = "Watchlist"
+        navigationController?.navigationBar.titleTextAttributes = [
+            .foregroundColor: UIColor.white,
+            .font: UIFont.systemFont(ofSize: 20, weight: .heavy)
+        ]
+        navigationController?.navigationBar.barTintColor = .darkBlue01
+    }
+   
+//    I've replaced viewDidLoad with viewDidAppear to ensure watchlist is up-to-date with the changes I make in other tabs. I guess I'll have to do the same for other screen so that watchlistButton appearance is up-to-date too.
```
GMinucci: Yup, exactly!

Features/WatchList/WatchlistViewController.swift
```diff
@@ -0,0 +1,104 @@
...
+                    }
+                case .failure(let error):
+                    DispatchQueue.main.async {
+                        ErrorViewController.handleError(error, presentingViewController: self)
+                    }
+                }
+            }
+        }
+    }
+}
```
GMinucci: Let's discuss this code in our sync

Features/WatchList/WatchlistViewController.swift
```diff
@@ -0,0 +1,104 @@
...
+
+// MARK: - WatchlistButtonDelegate
+extension WatchlistViewController: WatchlistButtonDelegate {
+//    It might make sense to add slightly different logic to this function. Like immediately deleting the row where the button was tapped.
```
GMinucci: Yep, what about refreshing the view content? That would take care of removing the row no?

NSLayoutConstraint.swift
```diff
@@ -0,0 +1,20 @@
...
+//  Movies
+//
+//  Created by Diana Chizhik on 1.08.22.
+//
+
+import Foundation
+import UIKit
+
+extension NSLayoutConstraint {
+    static func pin(_ subview: UIView, to superview: UIView, withInsets insets: UIEdgeInsets) {
```
GMinucci: I really liked this one! 🥇 
I have a couple suggestions to improve it too:
- Ensure that the subview is a child of the superview (otherwise this would crash)
- Ensure that `translatesAutoresizingMaskIntoConstraints` is false (you can maybe set it here so we don't need to do it everywhere)
- Bonus: Can we perhaps choose which corners to pin? In case I want to just pin top corner for example

UIColor.swift
```diff
@@ -0,0 +1,20 @@
...
+extension UIColor {
+    static let lightBlue61 = UIColor(named: "lightBlue61") ?? .blue
+    static let green = UIColor(named: "green") ?? .green
+    static let whiteF5 = UIColor(named: "whiteF5") ?? .white
+    static let darkBlue01 = UIColor(named: "darkBlue01") ?? .black
+    static let pureWhiteFF = UIColor(named: "pureWhiteFF") ?? .white
+    static let blue2A = UIColor(named: "blue2A") ?? .blue
+    static let lightBlueDB = UIColor(named: "lightBlueDB") ?? .lightGray
+    static let lightGreyB8 = UIColor(named: "lightGreyB8") ?? .lightGray
+    static let darkGrey43 = UIColor(named: "darkGrey43") ?? .lightGray
```
GMinucci, line 19: Usually the designers define the color name so it's easier for us, usually semantic color names like "warning" or "primary". Of course, that's not always the case haha what I like to do is to use this tool [here](https://www.color-name.com/) to get nicely named colors. Feel free to use if you want.

Views/WatchlistButton.swift
```diff
@@ -0,0 +1,80 @@
...
+import UIKit
+
+protocol WatchlistButtonDelegate: AnyObject {
+    func watchlistTapped(_ view: WatchlistHandleable)
+}
+
+protocol WatchlistHandleable: AnyObject {
+    var watchlistButtonDelegate: WatchlistButtonDelegate? { get set }
+    var watchlistButton: WatchlistButton { get }
+}
```
GMinucci, line 17: Our protocol implies that to be a `WatchlistHandleable` a class needs to have a button. What if it's simply a cell and you have a swipe action to remove from watchlist instead of a button?
Ideally the protocol should have the minimum required properties/methods to achieve the behaviour we need. And we need to be careful to no make the protocol forcing something we don't need. In this case, to be able to handle watchlist status we simply need something capable of changing its watchlist state, so:

```swift
protocol WatchlistHandleable {
    func didUpdateWatchlistStatus(to newStatus: WatchlistStatus)
}
```

Unless this protocol is only about views that can handle this button. If that's the case I'd suggest update the naming to maybe `WatchlistButtonHandleable`. And in that case you can also only enforce the methods, not the property:

```swift
protocol WatchlistButtonHandleable {
    func shouldUpdateWatchlistItem(with status: WatchlistStatus)
}
```

Services/WatchlistService.swift
```diff
@@ -0,0 +1,86 @@
...
+    func getWatchlistItems() -> [WatchlistItem]?
+}
+
+enum WatchlistStatus {
+    case added, notAdded
+}
+
+class WatchlistService {
+    typealias Watchlist = [Int : WatchlistItem]
+    static var watchlistKey: String = "watchlist"
```
GMinucci: A static `var` still allows for modification, in this case we want to avoid that so let's make it `static let`. To understand how this behaves:
```swift
struct Vehicle {
    static var subtype = "car"
    
    let name: String
    
    func printSubtype() {
        print(Self.subtype)
    }
}

let a = Vehicle(name: "Yaris")
let b = Vehicle(name: "Corsa")
a.printSubtype() // prints: "car"
b.printSubtype() // prints: "car"
print(Vehicle.subtype) // prints: "car"

Vehicle.subtype = "not a car"
a.printSubtype() // prints: "not a car"
b.printSubtype() // prints: "not a car"
print(Vehicle.subtype) // prints: "not a car"
```

GMinucci (changes requested):
Great work here! I tested the app and it's working like charm ❤️ 
I've added a couple minor comments and suggestions and on one topic I wanted to discuss in our sync.

Some general considerations (not for this PR but for future ones):
1. Let's try to break down PR size. 2k lines is a lot to review and can make it hard to properly review it, we could break it in smaller PRs, taking this for example: One to make only the watchlist service, one to integrate to UI, a separate one to refactor old code. You can always make a branch from another branch so you can keep working on and later just rebase before each merge.
2. Let's refactor the view controllers to be unknowledgeable about which service implementation they are using: We replace the variable type to the service protocol and initialize it through the init method. (with this, we have a dependency inversion fully in place). Of course, in a later PR.
3. Let's include the `.DS_Store` files to our git ignore, they are annoying haha

Features/PlayingNow/PlayingNowCollectionViewController.swift
```diff
@@ -8,12 +8,28 @@
...
-    private let movieDataService = MovieDataService()
-    private let watchlistService = WatchlistService()
+    private var movieDataService: MovieDataServiceProtocol
+    private var watchlistService: WatchlistServiceProtocol
     
+    private var movies = [Movie]()
     var selectedItemId: Int?
     
+//    Did I get you right when you talked about replacing services with protocols?
+//    I've made changes only on PlayingNoewVC so far.
```
GMinucci, line 18: Yep, you got it right, that's it 😄
If you want to read more about it, when you set a dependency (the services in this case) using the init it's called `Constructor injection`. There are several ways to achieve it like `Property injection`, `Method injection` or `Service locator`. I can explain them if you want as well.

GMinucci (approved):

PR #8, opened 2022-08-12
Core data

DiChizhik:


Features/DetailView/MovieDetailViewController.swift
```diff
@@ -8,23 +8,34 @@
 import UIKit
 import Kingfisher
 
+protocol MovieDetailViewDelegate: AnyObject {
+    func updateView(_ controller: UIViewController)
```
GMinucci: This method implies more what would be an interface to communicate than a delegate. Meaning that we are "ordering an action from another" than "delegating an action to another".
I would recommend to use something like `didUpdateWatchlist(_ viewController)` in the delegate and then this view does whatever it wants, that can be updating the whole view, just one item and so on.

Features/DetailView/MovieDetailViewController.swift
```diff
@@ -106,22 +126,22 @@ private extension MovieDetailViewController {
     }
     
     @objc func dismissView(_ sender: UIButton) {
+        delegate?.updateView(self)
```
GMinucci: Since this is being called on `viewWillDisappear` we don't need it here too

Features/DetailView/MovieDetailViewController.swift
```diff
@@ -106,22 +126,22 @@ private extension MovieDetailViewController {
...
-                                          voteAverage: movieDetails.voteAverage,
-                                          posterPath: movieDetails.posterPath)
+        guard let movie = movieDetails else { return }
 
-        let updatedStatus = watchlistService.toggleStatus(for: watchlistItem)
+        let updatedStatus = watchlistService.toggleStatus(for: WatchlistMovieConfiguration(id: Int32(movie.id),
+                                                                                           saveDate: Date.now,
+                                                                                           title: movie.title,
+                                                                                           voteAverage: Int16(movie.voteAverage),
+                                                                                           posterPath: movie.posterPath))
```
GMinucci: Can we add an initializer to `WatchlistMovieConfiguration` that receives an instance of `MovieDetails` as input so you can easily generate them and keep this logic there as well (casting `movie.id` to an Int32, setting date as `Date.now`, and casting vote average to an Int16)

Features/MostPopular/MostPopularViewController.swift
```diff
@@ -14,10 +14,11 @@ class MostPopularViewController: UIViewController {
...
     
-    var movies = [Movie]()
-    let movieDataService = MovieDataService()
-    let watchlistService = WatchlistService()
-    var itemInViewIndex: Int = 0
+    private let movieDataService: MovieDataServiceProtocol
+    private let watchlistService: WatchlistServiceProtocol
+    
+    private var movies = [Movie]()
+    private var itemInViewIndex: Int = 0
```
GMinucci, line 21: Access control  ❤️

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
 //
 
 import UIKit
+import CoreData
 
 protocol WatchlistServiceProtocol {
-    func getStatus(for id: Int)-> WatchlistStatus
-    func toggleStatus(for item: WatchlistItem) -> WatchlistStatus
-    func getWatchlistItems() -> [WatchlistItem]
+    func getStatus(for id: Int) throws -> WatchlistStatus
```
GMinucci, line 12: Some fancy throwing for errors, nice 👍

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
 }
 
 enum WatchlistStatus {
     case added, notAdded
 }
 
-class WatchlistService: WatchlistServiceProtocol {
-    typealias Watchlist = [Int : WatchlistItem]
-    private static let watchlistKey: String = "watchlist"
+enum WatchlistServiceError: Error, ErrorViewHandleable {
```
GMinucci, line 21: That's a great approach here: Creating a custom error type for this service and already taking advantage of our `ErrorViewHandleable` protocol. Amazing work! 100 points to Gryffindor 🦉

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+            }
+        }
+        return container
+    }()
     
-    func getWatchlistItems() -> [WatchlistItem] {
-        let watchlist = getWatchlist()
+    private init() {
+    }
+}
```
GMinucci: In my old days we usually had the persistent container declared in the AppDelegate, as AppDelegate is already a singleton you save a tiny bit of memory which... meh, I like this approach more. The only tiny change I'd suggest is that, for now, no one needs to know or access our Container so we could make this class private 😄

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+
+class WatchlistService: WatchlistServiceProtocol {
```
GMinucci, line 55: I'm not saying that it's wrong here but, when I said to implement the watchlist using UserPreferences first then CoreData after I was referring to in the end have 2 different classes, one using CoreData and one using UserPreferences so you could switch between them easilly by using the ServiceProtocol. 
It's quite common to have several implementations for the same service in a project, for example if you want to implement UnitTests you won't be using your real service that fetches from an API (imagine having 100 tests running every time you hit "test"... more than 100 API request at once), so in this case you'd have another service that fetches from a JSON locally, saving your internet data and making it faster (your backend and devops team will also thank you)
Long story short, no need to go back and add the old service again and bla bla bla... Just wanted you to be aware of this practice 😄

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+    let container = Container.shared
```
GMinucci: Can we make it private?

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+    var predicate: NSComparisonPredicate?
```
GMinucci: Do we need to keep an predicate reference alive all the time? Can we keep it only in the scope of the methods or is there any advantage of making it a property?

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+    func getStatus(for id: Int) throws -> WatchlistStatus {
+        let request = WatchlistMovie.createFetchRequest()
+        predicate = NSComparisonPredicate(format: "id == %@", id as NSNumber)
+        request.predicate = predicate
         
-        var items = [WatchlistItem]()
-        for value in watchlist.values {
-            items.append(value)
+        do {
+            let movies = try container.viewContext.fetch(request)
```
GMinucci: Since we are only interested in the count of results we can use. No big deal, it's just a small performance improvement.
```suggestion
            let movies = try container.viewContext.count(for: request)
```

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+            
+            if !movies.isEmpty {
+                return .added
+            } else {
+                return .notAdded
+            }
+        } catch {
+            print("Fetch failed")
```
GMinucci: Don't forget to remove prints after we are done debugging
```suggestion
```

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+    }
+}
+
+class WatchlistService: WatchlistServiceProtocol {
+    let container = Container.shared
+    var predicate: NSComparisonPredicate?
+
+    func getStatus(for id: Int) throws -> WatchlistStatus {
+        let request = WatchlistMovie.createFetchRequest()
+        predicate = NSComparisonPredicate(format: "id == %@", id as NSNumber)
```
GMinucci: I'm not sure on this one, but if you use like this, does it work?
```suggestion
        predicate = NSComparisonPredicate(format: "id == %d", id)
```

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
         
-        if let watchlistData = defaults.object(forKey: WatchlistService.watchlistKey) as? Data {
-            let decoder = JSONDecoder()
-            if let watchlist = try? decoder.decode(Watchlist.self, from: watchlistData) {
-                return watchlist
+        do {
+            let movies = try container.viewContext.fetch(request)
+            
+            if !movies.isEmpty {
+                container.viewContext.delete(movies[0])
```
GMinucci: Just another way to write the same code without accessing the array directly (nothing wrong with your code!)
```suggestion
            if let movie = movies.first {
                container.viewContext.delete(movie)
```

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+                return saveWatchlist() ? .notAdded : .added
             } else {
-                return Watchlist()
+                let watchlistMovie = WatchlistMovie(context: container.viewContext)
+                watchlistMovie.id = item.id
+                watchlistMovie.saveDate = item.saveDate
+                watchlistMovie.title = item.title
+                watchlistMovie.voteAverage = item.voteAverage
+                watchlistMovie.posterPath = item.posterPath
```
GMinucci: We are only setting this in a single place on our app but in case you have to do this in several places we can maybe consider a method in `WatchlistMovie` called maybe `configure(with movieDetails: MovieDetails)` so it makes this a bit easier.

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+
+                return saveWatchlist() ? .added : .notAdded
             }
-        } else {
-            return Watchlist()
+        } catch {
+//            replace with throwing function later
```
GMinucci, line 93: Sounds like a good idea indeed 😄

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+            return saveWatchlist() ? .added : .notAdded
         }
+        
     }
     
-    private func saveWatchlist(_ list: Watchlist) -> Bool {
-        let encoder = JSONEncoder()
+    func getWatchlist() -> [WatchlistMovie] {
+        let request = WatchlistMovie.createFetchRequest()
+        let sort = NSSortDescriptor(key: "saveDate", ascending: false)
```
GMinucci, line 101: Nice one! 👍

Services/WatchlistService.swift
```diff
@@ -6,76 +6,129 @@
...
+                do {
+                    try container.viewContext.save()
+                    return true
+                } catch {
+                    print("An error occurred while saving: \(error)")
+                    return false
+                }
+        }
+        
+        return true
```
GMinucci: Since `hasChanges` is a required step for this method, making it a `guard` improves readability
```suggestion
        guard container.viewContext.hasChanges else { return true }
        
        do {
            try container.viewContext.save()
            return true
        catch {
            print("An error occurred while saving: \(error)")
            return false
        }
```

WatchlistMovie+CoreDataProperties.swift
```diff
@@ -0,0 +1,36 @@
...
+//  Movies
+//
+//  Created by Diana Chizhik on 10/08/2022.
+//
+//
+
+import Foundation
+import CoreData
+
+struct WatchlistMovieConfiguration {
```
GMinucci, line 19: This may be a bit of personal opinion but I'm not sure about this naming choice, to be a "Configuration" type. I'd rather go to something like `WatchlistMovieModel` or simply `WatchlistMovie` and having the core data entity as `CDWatchlistMovie`.

GMinucci (commented):
Some small comments but overall another great PR! Congrats 🥇

Features/DetailView/MovieDetailViewController.swift
DiChizhik: Totally makes sense!

Features/DetailView/MovieDetailViewController.swift
DiChizhik: Right! Made changes to viewWillDisappear to fix a bug (that my personal QA Engineer found) and forgot to update the rest

Features/DetailView/MovieDetailViewController.swift
DiChizhik: Since `toggleStatus(for item: WatchlistMovieConfiguration)` function has to accept not only `MovieDetails`, but also `Movie` instances, I've created `CoreDataHandleable` protocol (not sure about naming as usual) to make one initializer for both and for possible new data models. And a separate initializer for `WatchlistMovie` instances.

Services/WatchlistService.swift
DiChizhik: It'd be great to discuss it in the call

Services/WatchlistService.swift
DiChizhik: What if I'm from Slytherin? :)

Services/WatchlistService.swift
DiChizhik: It's still done this way if you tick UseCoreData when creating a project. Though I find it much more transparent this way.

Btw, `private access restricts the use of an entity to the enclosing declaration, and to extensions of that declaration that are in the same file. ` What will be the enclosing declaration for `Container` class here? `WatchlistService `does need access to it, but it seems happy even though I changed access level to "private". Is `Container` now limited by the file?

Services/WatchlistService.swift
DiChizhik: OMG
I think I will go back and return UserDefaults :) It never hurts to have an alternative
I mean it doesn't take much effort, right?

Services/WatchlistService.swift
DiChizhik: I'm quite sure we can

Services/WatchlistService.swift
DiChizhik: You're right, it's redundant.

Services/WatchlistService.swift
DiChizhik: Great, thanks!

Services/WatchlistService.swift
DiChizhik: How do I handle an error here then? I mean if I delete prints, there will be no indicator that a fetch failed. However, if it did fail it's none of a user's business. So, it might be useful while testing, perhaps...

Services/WatchlistService.swift
DiChizhik: it was something that gave me a hard time with CoreData - the predicate.
it seems to work only in the variant I implemented.

Services/WatchlistService.swift
DiChizhik: Even for this one place it will look much neater with a separate method for configuring

WatchlistMovie+CoreDataProperties.swift
DiChizhik: If I'm asked what my weaknesses are, I'll say naming :)
Well, I don't think changing the name of CoreData entity is a good idea. I think it has once already ruined everything.
Let's discuss it in the call, shall we?

PR #9, opened 2022-08-19
Core data v2

DiChizhik:


Features/DetailView/MovieDetailViewController.swift
```diff
@@ -125,12 +125,15 @@ private extension MovieDetailViewController {
...
         
-        if let status = try? watchlistService.getStatus(for: selectedMovieID) {
-            contentView.watchlistButton.updateWithStatus(status, isShortVariant: false)
-        } else {
-            ErrorViewController.handleError(WatchlistServiceError.failedToFetchFromPersistentStore, presentingViewController: self)
+        watchlistService.getStatus(for: selectedMovieID) { [weak contentView] result in
+            switch result {
+            case .success(let status):
+                contentView?.watchlistButton.updateWithStatus(status, isShortVariant: false)
+            case .failure(_):
```
GMinucci, line 132: Just a note, when you don't need the associated value from a enum in a switch case you can simply omit it:
```swift
switch result {
case .success(let status):
    // ...
case .failure:
    // ...
```

Features/MostPopular/MostPopularViewController.swift
```diff
@@ -129,10 +135,13 @@ private extension MostPopularViewController {
             let movieDescription = movie.overview
             let movieID = movie.id
             
-            if let status = try? watchlistService.getStatus(for: movieID) {
-                contentView.configureWith(name: movieName, score: reviewsScore, description: movieDescription, status: status)
-            } else {
-                ErrorViewController.handleError(WatchlistServiceError.failedToFetchFromPersistentStore, presentingViewController: self)
+            watchlistService.getStatus(for: movieID) { [weak contentView] result in
```
GMinucci: Just to note here: When you say `[weak contentView]` you are weakly capturing `contentView` property only, not `self`. If you need access anything besides `contentView` it would still create a retain cycle

Features/PlayingNow/PlayingNowCollectionViewController.swift
```diff
@@ -89,13 +89,22 @@ private extension PlayingNowCollectionViewController {
...
+            watchlistService.toggleStatus(for: movie) { [weak view] result in
+                switch result {
+                case .success(let updatedStatus):
+                    view?.watchlistButton.updateWithStatus(updatedStatus, isShortVariant: true)
+                case .failure(_):
+                    return
+                }
+            }
         }
+//        let error = NSError(domain: "test error message", code: -1, userInfo: nil)
```
GMinucci: Is this an unused piece of code left behind? 👀

Features/WatchList/WatchlistViewController.swift
```diff
@@ -68,8 +78,7 @@ extension WatchlistViewController: WatchlistButtonDelegate {
         guard let indexPath = contentView.tableView.indexPath(for: cell) else { return }
         
         if let movie = movies[safe: indexPath.row] {
```
GMinucci: This is not from this PR but I just noticed, `movie` is a prerequisite for the method to run so it should be part of the guard statement at the beginning

Services/WatchlistService.swift
```diff
@@ -52,28 +54,38 @@ private class Container {
...
             if movies != 0 {
-                return .added
+                completion(.success(.added))
             } else {
-                return .notAdded
+                completion(.success(.notAdded))
             }
         } catch {
-            throw WatchlistServiceError.failedToFetchFromPersistentStore
+//            I'm not sure about the form of recording errors. This is what I've come up with so far. Tried recodring just WatchlistServiceError.failedToFetchFromPersistentStore. Looked ok to me as well, however NSError allows for sending additional info.
```
GMinucci: Another way here is to add a computed property to your error:
```swift
enum MyError: Int, Error {
    case failedToDoSomething
    
    var asNSError: NSError {
        NSError(domain: "MyErrorDomain", code: rawValue, userInfo: nil)
    }
}
```
and use here as:
```swift
Crashlytics.crashlytics().record(error: MyError.failedToDoSomething.asNSError)
```

Services/WatchlistService.swift
```diff
@@ -82,42 +94,168 @@ class WatchlistService: WatchlistServiceProtocol {
...
         } catch {
-            print("An error occurred while saving: \(error)")
-            return false
+//            report an error
+            completion(.failure(.failedToSaveToPersistentStore))
+        }
+    }
+}
+
+class UserDefaultsWatchlistService: WatchlistServiceProtocol {
```
GMinucci: Let's move this to its own dedicated class 😁

GMinucci (changes requested):
Not much to comment here, looks good, just a minor file organization

Features/PlayingNow/PlayingNowCollectionViewController.swift
DiChizhik: Not any longer :)

